PR #12, opened 2015-03-17
Fix code generator for double and float fields with [default=nan]

chrischamberlin:
nan is a valid default value, but ScalaPB generates code that doesn't compile if it is used.

Thanks for a great tool.


thesamet:
This looks great - thanks!

PR #23, opened 2015-05-08
build: upgrade scala versions

ahjohannessen:


thesamet:
I think the travis failure is a temporary error (things get a little slower on their end). Merging and will keep an eye on this. Thanks for the PR.

PR #45, opened 2015-10-20
update dependencies

xuwei-k:


eiennohito:
I believe, that there was Scala 2.12-M3 release as well couple of days ago.

xuwei-k:
> there was Scala 2.12-M3 release as well couple of days ago.

You're right, but some dependency libraries does not published yet with Scala 2.12.0-M3.

thesamet:
Thank you!

PR #52, opened 2015-11-06
Target JDK 1.7

plaflamme:
The codebase is not using any JDK 1.8 features. This ensures that when compiling with JDK 1.8, the artifacts produced are compatible with 1.7.


plaflamme:
Not sure why the build is failing. Looks like maybe it's timing out?

> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

Testing locally works fine.

thesamet:
Yep, it was just time outs. We have been seeing it before. The randomly generated protocol buffers sometimes take really long time to compile on Travis CI.

PR #53, opened 2015-11-07
target java7

xuwei-k:
https://github.com/trueaccord/ScalaPB/pull/52 is not enough because

- there are some java files in ScalaPB
- does not support `-target:jvm-1.7` option Scala 2.12.0-M2 or later


thesamet:
Thanks for fixing this!

PR #63, opened 2015-12-16
use fully qualified name. s/Any/scala.Any

xuwei-k:
I want to use `google/protobuf/any.proto`

https://github.com/google/protobuf/blob/v3.0.0-beta-1/src/google/protobuf/any.proto

```
[error] target/scala-2.11/src_managed/main/compiled_protobuf/com/google/protobuf/any/Any.scala:60: type mismatch;
[error]  found   : String
[error]  required: com.google.protobuf.any.Any
[error]         case 1 => typeUrl
[error]                   ^
[error] target/scala-2.11/src_managed/main/compiled_protobuf/com/google/protobuf/any/Any.scala:61: type mismatch;
[error]  found   : com.google.protobuf.ByteString
[error]  required: com.google.protobuf.any.Any
[error]         case 2 => value
[error]                   ^
[error] two errors found
```

### DIFF
--- a/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
@@ -246,7 +246,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
     }
 
   def generateGetField(message: Descriptor)(fp: FunctionalPrinter) = {
-    val signature = "def getField(__field: com.google.protobuf.Descriptors.FieldDescriptor): Any = "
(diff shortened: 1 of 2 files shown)


xuwei-k:
> LG, can you add an Any message to e2e/src/main/protobuf/names.dot.proto ?

done

thesamet:
Thank you!

zackangelo:
@thesamet @xuwei-k There's a similar namespace conflict with the `Option` message, any objections to a PR that fully qualifies reference to Scala's `Option` class as well? (see https://github.com/google/protobuf/blob/v3.0.0-beta-1/src/google/protobuf/type.proto#L163)

nadavsr:
No objection, PR will be welcome.

PR #72, opened 2016-01-07
Fix double stream close, don't call onComplete after onError

### LINKED ISSUE #71: gRPC: Should not call onComplete after onError
We're calling `onComplete` after `onError` when converting a call's result from a `Try` to a `StreamObserver`. This violates the interface contract of `StreamObserver`, which says `onError` should be the last method called: https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/StreamObserver.java#L63 

This causes the stream to be double-closed, resulting in a `IllegalStateException` when an error occurs inside of a gRPC service method.

zackangelo:
Fixes #71


thesamet:
Thanks for fixing this

PR #96, opened 2016-03-23
fix trueaccord/ScalaPB#94

cheleb:
Camel casing issue: Keep underscore when it is separating two digits.
So x1_1x is no more conflicting with x11x when camelcased.
Add openrtb.proto from https://github.com/google/openrtb


thesamet:
Underlying issue solved - please let me know if additional changes are necessary.

Issue #108, opened 2016-05-04
Make grpc service name visible in generated code


eiennohito:
Right now it is inlined, but exposing it will make it easier for more advanced code.

For example, I am trying to use grpc with akka streams for backpressure support.

I can submit a pull request containing base trait for services and patch for generator.

`trait GrpcService { def name: String }`

thesamet:
Fixed by a754b35a226fba283acf35a5245de8617272aec6.

The service name can be exposed through Service.descriptor.getName()

PR #112, opened 2016-05-08
expose grpc service name as a val

eiennohito:
solution for #108


thesamet:
Thanks for working on this. It's preferable to expose the entire ServiceDescriptor (which includes the service name among other things)

eiennohito:
Yes, I agree with you. Thank you for the implementation.

PR #114, opened 2016-05-08
depend not on grpc-all, but on more fine-grained dependencies

eiennohito:
fixes #113 

Need to fix docs as well by introducing explicit dependency on a transport, otherwise default examples will stop working.

PR #115, opened 2016-05-08
update documentation for removing grpc-all

eiennohito:
Fixes documentation to be aligned with #114 

Fixes a problem in example as well.


thesamet:
Thanks!

Issue #116, opened 2016-05-11
Performance issue tracking


eiennohito:
It takes a lot in profiles. 3-4 times more than actual serialization of the message (writeTo).

However, before doing anything need to write a benchmark for this.

thesamet:
Thanks for filing this. At this point, we haven't done much to optimize the performance of ScalaPB but it's been on my mind. It would be really nice to have an example benchmark we can work with.

eiennohito:
I will try to do benchmarks and to experiment making `serializedSize` a simple (non-volatile) var with a sentinel value like java.lang.String does with its hashcode. 

But not right now.

thesamet:
Can you share some details on the message that caused the slow serializedSize time? Deeply nested? Big arrays? Maps? One ofs? Anything else?

eiennohito:
I don't know yet because I did not do exact benchmarking.
I believe that lazy val is one of main culprits.

eiennohito:
Okay, I've done some benchmarking.

Project: https://github.com/eiennohito/scalapb-jmh-bench
[Benchmarking Code](https://github.com/eiennohito/scalapb-jmh-bench/blob/master/src/main/scala/org/eiennohito/Bench2.scala), [Protobuf](https://github.com/eiennohito/scalapb-jmh-bench/blob/master/src/main/protobuf/test.proto)

For the test data I use three level nesting: `Obj1[ 10x Obj2 [ 50x Obj3[ string ] ] ]`
There are three setups:
* copyOnly (baseline): doing a deep copy (because we need to reinit `serializedSize` of all objects)
* copyAndSize: baseline + calculation of serializedSize
* copyAndSerialize: copyAndSize + creating `CodedOutputStream` with preallocated array + serialization of size + serialization of the message itself

I run tests on JDK 1.9.0_92, Xeon E5-2667 v3, CentOS 6.7

```
Benchmark                Mode  Cnt   Score   Error  Units
Bench2.copyAndSerialize  avgt   10  31.266 ± 0.180  us/op
Bench2.copyAndSize       avgt   10  23.249 ± 0.174  us/op
Bench2.copyOnly          avgt   10   4.002 ± 0.006  us/op
```

Basically there is +20microsec/op on calculating size and +10microsec/op on serializing the object itself. I will use this project for the further analysis.

eiennohito:
Benchmarking serializedSize
===========================

Methodology:
* Using JMH
[...]

PR #119, opened 2016-05-18
replace lazy vals in size generation with explicit lazy initialization

eiennohito:
lazy vals are replaced by transient backing fields with -1 as sentinel value
 and lazy initialization in accessor methods

 The best low hanging fruit for #116 right now.

 Motivation for the fix is described in the same #116


thesamet:
Merged - thank you for looking into this.

PR #122, opened 2016-06-01
fixes #121: java serialization breaks serializedSize

### LINKED ISSUE #121: [0.5.27 regression] serializedSize does not work with java serialization
Do not use 0.5.27, 0.5.28 if you serialize case classes with java serialization and then serialize them as protobufs.

I am working on the fix.

eiennohito:
Previously -1 was used as a sentinel value.
 However, if a generated case class to be serialized using java serialization,
 all backing fields would be reset to 0 because java serialization does not call
 constructors.

 This fix changes serializedSize default (sentinel) value to 0.
 Serialized size of 0 is valid value and it will not be cached if the message
 is actually empty.
 Still, for such messages computing size should be cheap.

### DIFF
--- a/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
@@ -372,7 +372,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
   def generateSerializedSize(message: Descriptor)(fp: FunctionalPrinter) = {
     fp
       .add("@transient")
-      .add("private[this] var __serializedSizeCachedValue: Int = -1")
+      .add("private[this] var __serializedSizeCachedValue: Int = 0")
       .add("private[this] def __computeSerializedValue(): Int = {")
       .indent
       .add("var __size = 0")
@@ -382,8 +382,14 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       .add("}")
       .add("final override def serializedSize: Int = {")
       .indent
-      .add("if (__serializedSizeCachedValue == -1) { __serializedSizeCachedValue = __computeSerializedValue() }")
-      .add("__serializedSizeCachedValue")
+      .add("var read = __serializedSizeCachedValue")
+      .add("if (read == 0) {")
+      .indent
+      .add("read = __computeSerializedValue()")
+      .add("__serializedSizeCachedValue = read")
+      .outdent
+      .add("}")
+      .add("read")
       .outdent
       .add("}")
   }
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -382,8 +382,14 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       .add("}")
       .add("final override def serializedSize: Int = {")
       .indent
-      .add("if (__serializedSizeCachedValue == -1) { __serializedSizeCachedValue = __computeSerializedValue() }")
-      .add("__serializedSizeCachedValue")
+      .add("var read = __serializedSizeCachedValue")
```
thesamet: can you get rid of 'read' and use '__serializedSizeCachedValue` directly instead?

eiennohito: It's two field accesses vs one.

JDK's toString hashCode and protobuf's ByteString use similar style:
https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/ByteString.java#L842

But I can make it back if you still want.

thesamet: Got it, no need to change. Will merge in.

On Wed, Jun 1, 2016 at 6:32 AM, Arseny Tolmachev <notifications@github.com>
wrote:

> In
> compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
> <https://github.com/trueaccord/ScalaPB/pull/122#discussion_r65359890>:
>
> > @@ -382,8 +382,14 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
> >        .add("}")
> >        .add("final override def serializedSize: Int = {")
> >        .indent
> > -      .add("if (__serializedSizeCachedValue == -1) { __serializedSizeCachedValue = __computeSerializedValue() }")
> > -      .add("__serializedSizeCachedValue")
> > +      .add("var read = __serializedSizeCachedValue")
>
> It's two field accesses vs one.
>
> JDK's toString hashCode and protobuf's ByteString use similar style:
>
> https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/ByteString.java#L842
>
> But I can make it back if you still want.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/trueaccord/ScalaPB/pull/122/files/b8d6a4371ba8ce91dee94067ec51feb798b00275#r65359890>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AASwrfRPDoFIScpVxSw_2Wp41EZEaqzCks5qHYoHgaJpZM4Irdnj>
> .
>



-- 
-Nadav

PR #157, opened 2016-08-03
use forNumber instead of valueOf. fix deprecation warnings

xuwei-k:
https://github.com/google/protobuf/blob/v3.0.0/src/google/protobuf/compiler/java/java_enum.cc#L183-L189


thesamet:
Thanks

PR #161, opened 2016-08-12
port ProtoUtils.keyForProto from grpc-java

chikei:
Utility methods for putting protos into `Metadata`

### DIFF
--- a/scalapb-runtime-grpc/src/main/scala/com/trueaccord/scalapb/grpc/ProtoUtils.scala
+++ b/scalapb-runtime-grpc/src/main/scala/com/trueaccord/scalapb/grpc/ProtoUtils.scala
@@ -1,17 +1,17 @@
 package com.trueaccord.scalapb.grpc
 
 import com.google.protobuf.InvalidProtocolBufferException
-import com.trueaccord.scalapb.GeneratedMessage
+import com.trueaccord.scalapb.{GeneratedMessage, GeneratedMessageCompanion, Message}
 import io.grpc.Metadata
 
 object ProtoUtils {
-  def metadataMarshaller[T <: GeneratedMessage](instance: T): Metadata.BinaryMarshaller[T] =
+  def metadataMarshaller[T <: GeneratedMessage with Message[T]](implicit companion: GeneratedMessageCompanion[T]): Metadata.BinaryMarshaller[T] =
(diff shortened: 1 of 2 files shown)


thesamet:
@xuwei-k , can you review?

scalapb-runtime-grpc/src/main/scala/com/trueaccord/scalapb/grpc/ProtoUtils.scala
```diff
@@ -0,0 +1,25 @@
+package com.trueaccord.scalapb.grpc
+
+import com.google.protobuf.InvalidProtocolBufferException
+import com.trueaccord.scalapb.GeneratedMessage
+import io.grpc.Metadata
+
+object ProtoUtils {
+  def metadataMarshaller[T <: GeneratedMessage](instance: T): Metadata.BinaryMarshaller[T] =
```
eiennohito: You can solve this with typeclasses: `implicit companion: GeneratedMessageCompanion[T]`

I don't think it's a good idea to pass instance around. 

Even better, ScalaPbMetadataMarshaller should be an abstract class (because traits can't have parameters yet).

eiennohito:
I've added some comments if you are for this functionality.

Also, scala->java roundtrip tests are missing.

scalapb-runtime-grpc/src/main/scala/com/trueaccord/scalapb/grpc/ProtoUtils.scala
chikei: Will do using implicit companion.

About ScalaPbMetadataMarshaller, I am afraid I don't follow it, can you elaborate ?

eiennohito: I mean you create an anonymous class for the marshaller. Maybe (I'm not sure) it's better to use abstract class for the functionality so that it's possible to use that as `object` in other places or override something if the need comes. 
Then the `metadataMarshaller` method would just simply instantiate that.

chikei:
@eiennohito PTAL

eiennohito:
LGTM, but please somebody check else as well.

thesamet:
Squashed, minor edit and merged. Thank you @chikei .

PR #193, opened 2016-11-20
Added _root_ to generated message companion name to avoid conflicts with package names

### LINKED ISSUE #176: field names may conflict with package names
As reported in: 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/scalapb/peF0JdypMEk/Rhx4LAaaAwAJ

> If we define a message called "Msg" with an optional field called "org", which is, for instance, a int32, in a package "org.xx.xxx", then in the scalaPB-generated code (Msg.scala) there will be a field called "org" of type Option[Int]. However, there will also be a companion called "org.xx.xxx.PROTONAME.Msg", and there will be a line of code:
> 
> def companion = org.xx.xxx.PROTONAME.Msg
> 
> inside the case class "Msg" in the generated file. Since here "org" is already defined as a field, the complier is confused and thinks the "org" in "org.xx.xxx.PROTONAME.msg" is that Option[Int] field, which complains that "xx is not a member of Option[Int].

serejja:
Fixes #176


thesamet:
Thanks for working on this. There was another point in the code that needed changing and I made it also prepend _root_ only when needed.

serejja:
oh, cool! thanks! can you also publish an updated artifact so we can use it in the project?

thesamet:
Yes, would like to get a few more changes in today and will publish.

serejja:
Great, thank you for THAT quick reaction :)

PR #198, opened 2016-11-25
enable wrapper support by default

devzer01:
i don't expect you to merge this without a chat, but the way you have designed to enable wrappers using options breaks some C# based proto generators, because it doesn't agree with proto3/proto2 mix matches with descriptors. So there-for this was the quickest thing for me to do to get things running, doing this have removed the need to include scalapb.proto file and allows me to have wrapper enabled code. Thank you :)

### DIFF
--- a/compiler-plugin/src/main/java/com/trueaccord/scalapb/Scalapb.java
+++ b/compiler-plugin/src/main/java/com/trueaccord/scalapb/Scalapb.java
@@ -205,7 +205,7 @@ public final class Scalapb {
       import_ = com.google.protobuf.LazyStringArrayList.EMPTY;
       preamble_ = com.google.protobuf.LazyStringArrayList.EMPTY;
       singleFile_ = false;
-      primitiveWrappers_ = false;
+      primitiveWrappers_ = true;
     }
 
     @java.lang.Override


thesamet:
I was planning to have primitive wrapper wrapper enabled by default in a later release. I'm curious though what's the issue you are seeing with the C# based protos and the options. Can you give an example that demos this problem?

devzer01:
@thesamet i will try to get this info for you on Monday as i don't have them first hand, it was another person who was handling the c# side of the project i used ScalaPB, since i created this PR i spent lot of time reading Google protobuf docs and your code, 

making just your proto file proto3 compliant should also fix the problem in my opinion, unless they have used some custom compiler (which i hope not) the descriptor stuff should be part of the code already, so i think you don't actually need to include the reference to descriptor from scalapb file. will let you know as soon as i have info. 

Also i am trying to see if it's possible to override/extend just the compiler-plugin part to something that's suited for my needs. 

Thanks for replying quickly

thesamet:
Ok, so it looks like google's csharp code generator can only do proto3. What I suggest is that for C#, you set the import search path such that this version of scalapb.proto is picked up: https://gist.github.com/thesamet/d09eefdcb36efb8c918f62d70787a7f9

This file is exactly like the default scalapb.proto, but changed to proto3, so things should compile. Since these are just descriptor options, this shouldn't have any unintended consequence.

devzer01:
cool, thanks alot, will let you know how it goes, really appreciate your help, i was getting worried :) you saved my weekend :)

devzer01:
@thesamet on my end the new scalapb.proto3 file is working as expected, 
i will confirm Monday on c# compiler. 
also i experienced an issue with primitive strings. when data is loaded from database a database `NULL` value gets pass into the string and later on when accessing this string field `NullPointerExeption` if you believe this should be fixed from the compiler side let me know i can send a PR, i already  looked at the ScalaPB code quiet extensively 

If you believe this is not the responsibility of the compiler then no problem, i will wrap it on my end

thesamet:
The new proto should be used just when generating csharp code. please keep
using the original one when compiling for scala.

The null check should happen before passing it to ScalaPB. There isn't
anything generally useful that ScalaPB can do when getting detecting these
nulls -  it's just case classes so they're allowed

devzer01:
ok, not sure why need to keep the old one in Scala code, why not update the code? or add a build configuration flag for build.sbt to choose which version of the file need to be?

thesamet:
The reason to keep the new one out from your Scala build: if you generate code for it (rather than import it), the generated classes will not be source compatible with the ones shipped with ScalaPB, so you might run into issues at run time - however I think this will not happen if you name the file anything else but `scalapb.proto` (since the file name determines the Scala package name). Just need to be aware of it.

devzer01:
So what you are saying is that there is no plan to provide a proto3 compatible class options file for scalapb? 
Just as you pointed out the risks of not using the original file provided in ScalaPB, i am also concerned about using 2 different meta sources to generate client classes. So i will only use the latest one provided by you.

Thanks for your answer about Null values

thesamet:
@devzer01 yes, using only the latest one should be fine, if you rename it.

I can look into having scalapb.proto converted to proto3 in the future, don't want to break things for those who are stuck in the opposite direction (everything must be proto2).

jenol:
I just wanted to let you guys know that I am using wrappers.proto and the C# nullable generation works as expected: 

...
    google.protobuf.Int32Value contentFacilityGroupTypeId = 6;

...
    public int? ContentFacilityGroupTypeId {
      get { return contentFacilityGroupTypeId_; }
      set {
        contentFacilityGroupTypeId_ = value;
      }
    }

Thanks for all your hard work!

thesamet:
Primitive wrappers are now on by default on master.

PR #206, opened 2016-12-04
make generated companions extends FunctionN

xuwei-k:


thesamet:
What's the use case? Is it something everyone would need or better be a setting?

xuwei-k:
A good use case is as follows;
Suppose we have `val foo: Option[String]` and a generated message `case class Bar(value: String) extends GeneratedMessage`, we can just write `foo.map(Bar)` instead of writing `foo.map(Bar(_))`. This results in shorter code, and what's more, we don't have to create unnecessary function objects.

xuwei-k:
>  Is it something everyone would need or better be a setting?

I'll update to configurable.

xuwei-k:
> > Is it something everyone would need or better be a setting?
> 
> I'll update to configurable.

done

thesamet:
Seems like this would be mostly serving a style preference, since it's possible to write `foo.map(Bar.apply)` to avoid the creation of a function object. I would like to avoid the additional complexity. Closing the PR for now. I appreciate the time you took to write it, and will be happy to rethink this in the future if there's demand.

PR #214, opened 2016-12-14
Fix #213

### LINKED ISSUE #213: incorrect ByteString package name in generated sources in ScalaJs
an easy repro is to clone https://github.com/thesamet/scalapbjs-test and add a field to the example proto like:
```
syntax = "proto3";

package com.trueaccord;

import "scalapb/scalapb.proto";

message Food {
  string name = 1;
  string measure = 2;
  int32 calories = 3 [(scalapb.field).type = "com.trueaccord.Calories"];
  bytes something = 4;
}

message Menu {
  string menu_name = 1;
  repeated Food foods = 2;
}
```

result in compile error:
```
object com is not a member of package scalapb
[error]     something: scalapb.com.google.protobuf.ByteString = _root_.com.google.protobuf.ByteString.EMPTY
```
and subsequents.

andreaTP:
I haven't found what where but here an implicit conversion fix the string
```com.google.protobuf.ByteString``` to ```scalapb.com.google.protobuf.ByteString```
adding the ```_root_``` package first (as done in other places) preserve from this.

### DIFF
--- a/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/DescriptorPimps.scala
+++ b/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/DescriptorPimps.scala
@@ -182,7 +182,7 @@ trait DescriptorPimps {
         case "google.protobuf.FloatValue" => Some("Float")
         case "google.protobuf.StringValue" => Some("String")
         case "google.protobuf.BoolValue" => Some("Boolean")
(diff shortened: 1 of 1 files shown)


andreaTP:
thanks, when will be available a published release? any ETA?

thesamet:
Later this week. I'll ping the thread once that happens.

andreaTP:
thanks a lot!

Issue #208, opened 2016-12-07
Add option to generated repeated fields as Sets


Sushisource:
Mentioned this in the google group, adding it here.

It would be nice to have an option that can be set that would indicate a repeated field should appear as a Set in the generated code. Even better would be being able to specify anything that implements Iterable.

Thanks!

PR #218, opened 2016-12-23
add collection_type settings

xuwei-k:
https://gitter.im/trueaccord/ScalaPB?at=583dc53cb4d7ca3b7a1810fa

> Hello. Is there a way to generate List instead of Seq for repeated? I wrote a library that can convert between normal domain case classes and scalaPB generated case classes where all fields are optional. https://github.com/kailuowang/henkan#transform-between-case-classes-with-optional-field, however for repeated items, I need cats Functor and Traverse instances. There are no such instances for Seq provided out of the box.

/cc @kailuowang

Edit: also fix https://github.com/scalapb/ScalaPB/issues/208

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/package.scala
+++ b/compiler-plugin/src/main/scala/scalapb/package.scala
@@ -1,3 +1,4 @@
+import com.trueaccord.scalapb.compiler.ProtobufGenerator
 import protocbridge.JvmGenerator
 
 package object scalapb {
@@ -5,7 +6,8 @@ package object scalapb {
     flatPackage: Boolean = false,
     javaConversions: Boolean = false,
     grpc: Boolean = true,
-    singleLineToString: Boolean = false): (JvmGenerator, Seq[String]) =
+    singleLineToString: Boolean = false,
+    collectionType: String = "scala.collection.Seq"): (JvmGenerator, Seq[String]) =
     (JvmGenerator(
       "scala",
       ScalaPbCodeGenerator),
@@ -14,5 +16,6 @@ package object scalapb {
         "java_conversions" -> javaConversions,
         "grpc" -> grpc,
         "single_line_to_string" -> singleLineToString
-      ).collect { case (name, v) if v => name })
+      ).collect { case (name, v) if v => name } :+ (ProtobufGenerator.collectionTypeKey + "=" + collectionType)
+    )
 }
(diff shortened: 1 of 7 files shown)


kailuowang:
This is awesome! Thanks a lot! @xuwei-k

compiler-plugin/src/main/scala/scalapb/package.scala
```diff
@@ -1,11 +1,13 @@
...
 import protocbridge.JvmGenerator
 
 package object scalapb {
   def gen(
     flatPackage: Boolean = false,
     javaConversions: Boolean = false,
     grpc: Boolean = true,
-    singleLineToString: Boolean = false): (JvmGenerator, Seq[String]) =
+    singleLineToString: Boolean = false,
+    collectionType: String = "scala.collection.Seq"): (JvmGenerator, Seq[String]) =
```
thesamet: I prefer that this would be a field-level setting (in scalapb.proto), not a global compiler setting. The use case I see is that a single repeated field in a proto is known to be a set, while the rest of the file stays with the default implementation.

kailuowang: A field level set is definitely a good use case but for our projects being able to set a global collection type for repeated items different than `Seq` would be really helpful. `Seq` isnt very strict so that the functional programming libs we uses don't support it.

thesamet: Let's do a file-level option then. We found that global compiler settings (such as flat_package) that affect the public API of the generated code hurts maintainability issues down the line.

eiennohito: A set of three options would be the best in my opinion.
In some cases I would really like to use plain arrays of primitives for the performance reasons, for example.

So, options having priorities from high to low in the following order would be really the perfect variant:
Field level option -> File-level option -> Global option -> Seq

hseeberger:
What's the status of this PR? We urgently need to be able to use immutable sequences.

thesamet:
I looked into it today, started to make some changes and it's currently parking in the [collection_type](https://github.com/scalapb/ScalaPB/tree/collection_type). The main difference is that it's going to have a file and field-level options, but not as a global option. Global options should not affect the generated interfaces (I made that mistake with `flat_package` already) - it makes it impossible for other generators to know what interface to expect.

The current issues that still need to be solved:
- The lenses library has something called `SeqLens` that only works with `Seqs`. It needs to be parameterized over the collection type.
- `getField()` and `fromFieldMap` are used by TextFormat and the json conversion library. Those assume that `scala.collection.Seq` are passed for repeatables and they do `asInstanceOf[$collectionType]`, which would throw at runtime (for things like `Array` and `Vector`, but not `immutable.Seq`). Scala descriptors (coming soon) will have a type-safe version of `getField` and `fromFieldsMap` and TextFormat and JSON are ported to work with them. It will be a better time to potentially break these two methods.
- Minor: Some collection types won't work with Java conversions - `Array` doesn't have `asJava`, probably not a big deal.

thesamet:
Thanks for the PR @xuwei-k . Committed 5ce86e9 which is based on your work.

PR #250, opened 2017-03-10
Introduce getValue method to generated oneof traits for easier value access

Sushisource:
As discussed here:
https://groups.google.com/forum/#!topic/scalapb/CDkFfSS1Av4

Let me know if I need to make any updates to the PR. Thanks!

### DIFF
--- a/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
@@ -82,6 +82,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       .add(s"def isEmpty: Boolean = false")
       .add(s"def isDefined: Boolean = true")
       .add(s"def number: Int")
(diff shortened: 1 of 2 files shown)


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -82,6 +82,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       .add(s"def isEmpty: Boolean = false")
       .add(s"def isDefined: Boolean = true")
       .add(s"def number: Int")
+      .add(s"def getValue: Option[scala.Any] = None")
```
thesamet: Can you rename `getValue` to `valueOption` (like the standard `headOption` and `lastOption`), the reason is that in ScalaPB usually `x` returns an `Option[X]` and `getX` returns an `X`, so `getValue` would be inconsistent with that.

Sushisource: Sure thing, makes sense.

PR #264, opened 2017-04-04
Faster compilation of GeneratedOneof with a lot of members

avakhrenev:
Starting from `oneof` with number of members > 15 scalac spends significant amount of time in typer, trying to come up with pricise least upper bound types like

```scala
OneOf{def value: GeneratedMessage with Product with Message[_ >: OneOf.A with OneOf.B with ... <: GeneratedMessage with Product] with Updatable[_ >: OneOf.A with OneOf.B with ... <: GeneratedMessage with Product]{def companion: Serializable}}]
```

This PR frees scalac from this struggle by specifying the types explicitly.


thesamet:
I amended the commit such that the generated code has a chain of `orElse`'s rather than nested orElses. See  ff375c90f84d9d84359c2a392cf343de8f43bb0b let me know if there are any concerns on this change.

avakhrenev:
Great! Thank you.

PR #269, opened 2017-04-10
Fix typo in readme

aochagavia:


aochagavia:
@thesamet By the way, I saw that the Versions table is outdated.

PR #293, opened 2017-06-27
remove unnecessary anonymous function objects

xuwei-k:
We can omit `(_)` because PValue case class companion objects extend Function1

https://stackoverflow.com/q/3049368


thesamet:
Nice!

PR #302, opened 2017-07-11
fix #301

### LINKED ISSUE #301: Unused member __fields
It seems that Scala is generating a member `__fields` which is never used:

```
[warn] /data/rw/code/scala/zinc/internal/zinc-persist/target/scala-2.12/src_managed/main/sbt/internal/inc/schema/Public.scala:38:9: local val __fields in method fromFieldsMap is never used
[warn]     val __fields = javaDescriptor.getFields
[warn] 
```
I wonder if it's possible to make the codegen ignore the generation of this field.

xuwei-k:


thesamet:
Can you change the commit message to something along the lines of ```Remove unused variable `__fields` when it's empty to prevent a warning.

Fixes #301```

xuwei-k:
done

thesamet:
ty!

Issue #304, opened 2017-07-12
Support custom base traits for enums


ejconlon:
ScalaPB allows you to set "extends" and "companion_extends" to customize generated code for messages, but does not allow you to do the same for enums.

Practically speaking, my team noticed that scalacheck-shapeless was failing to generate Arbitrary instances for our generated enums, but we were unable to mix in a simple trait to remedy that.

Please let me know if you think this is a desirable feature, my team would be happy to help.  Thanks!

thesamet:
Yes, a PR that implements `extends` and `companion_extends` is welcome.

ejconlon:
Thanks.  Also, is there an objection to adding `type ValueType` to `GeneratedMessageCompanion` just like `GeneratedEnumCompanion`? It makes it easier to mix in things that refer to the message type.

thesamet:
Sounds good.

ruseinov:
@thesamet Hey Nadav, if you have any clues/preferences on how you want extends option for Enums implemented - please let me know. I did take a look at MessageOptions and a parser, it is not entirely straightforward, but doable.

thesamet:
Had some time to do this myself. It will be available in ScalaPB 0.6.1. The docs are in "Custom base traits for enums" in https://scalapb.github.io/customizations.html

ruseinov:
@thesamet got it

ejconlon:
Thanks @thesamet @ruseinov ! Is there a target date for the release?

thesamet:
Just released 0.6.1

PR #305, opened 2017-07-13
Introduced ValueType for messages

ruseinov:
As discussed in https://github.com/scalapb/ScalaPB/issues/304 with Eric I'm submitting a ValueType feature for MessageCompanions.

### DIFF
--- a/scalapb-runtime/shared/src/main/scala/com/trueaccord/scalapb/GeneratedMessageCompanion.scala
+++ b/scalapb-runtime/shared/src/main/scala/com/trueaccord/scalapb/GeneratedMessageCompanion.scala
@@ -173,6 +173,8 @@ trait JavaProtoSupport[ScalaPB, JavaPB] extends Any {
 }
 
 trait GeneratedMessageCompanion[A <: GeneratedMessage with Message[A]] {
+  type ValueType = A
+
   def parseFrom(input: CodedInputStream): A = LiteParser.parseFrom(this, input)
 
   def parseFrom(input: InputStream): A = parseFrom(CodedInputStream.newInstance(input))


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1074,6 +1074,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       s"""object $className extends $companionType {
          |  implicit def messageCompanion: $companionType = this""")
       .indent
+      .add(s"type ValueType = $className")
```
thesamet: Why not just type `type ValueType = A` in the  `GeneratedMessageCompanion` trait analogously to what's done for enums?

ruseinov: .add(s"type EnumType = $name")

ruseinov: Is it not the same thing? We get EnumType = ClassName and the being able to reference this type elsehwere. Correct me if I'm wrong.

thesamet: No, the difference is between adding this type to each individual instance versus adding it just once to the base trait. The latter has the advantage that generic code that works with the base trait will be able to access ValueType. Does that make sense?

ruseinov: Yep, it does.

I will change the implementation later on today.

ruseinov:
@thesamet Done!

PR #307, opened 2017-07-17
Better metadata for server service definitions.

dpratt:
Make the default wrapper create an io.grpc.ServerServiceDefinition with a proper io.grpc.ServiceDescriptor, which allows for the stock ProtoReflectionService to function properly.

### DIFF
--- a/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/GrpcServicePrinter.scala
@@ -175,6 +175,34 @@ final class GrpcServicePrinter(service: ServiceDescriptor, override val params: 
           |""")
   }
 
+  private[this] def serviceDescriptor(service: ServiceDescriptor) = {
+
+    val grpcServiceDescriptor = "_root_.io.grpc.ServiceDescriptor"
+
+    val schemaDescriptor: String =
+      s""".setSchemaDescriptor(new _root_.io.grpc.protobuf.ProtoFileDescriptorSupplier {
+         |  override def getFileDescriptor: _root_.com.google.protobuf.Descriptors.FileDescriptor =
+         |    ${service.getFile.fileDescriptorObjectFullName}.javaDescriptor
+         |})"""
+
+    def addMethod(method: MethodDescriptor): PrinterEndo = PrinterEndo(
+      _.add(s".addMethod(${method.descriptorName})")
+    )
+
+    PrinterEndo(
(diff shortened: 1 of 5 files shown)


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -175,6 +175,34 @@ final class GrpcServicePrinter(service: ServiceDescriptor, override val params:
           |""")
   }
 
+  private[this] def serviceDescriptor(service: ServiceDescriptor) = {
+
+    val grpcServiceDescriptor = "_root_.io.grpc.ServiceDescriptor"
+
+    val schemaDescriptor: String =
+      s""".setSchemaDescriptor(new _root_.io.grpc.protobuf.ProtoFileDescriptorSupplier {
```
thesamet: Instead of generating an anonymous supplier class for each service, I prefer that there would be a single class named `ConcreteProtoFileDescriptorSupplier` with a constructor that takes the descriptor and just return it. 

in scalapb-runt-ime-grpc

compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -175,6 +175,34 @@ final class GrpcServicePrinter(service: ServiceDescriptor, override val params:
...
+    def addMethod(method: MethodDescriptor): PrinterEndo = PrinterEndo(
+      _.add(s".addMethod(${method.descriptorName})")
+    )
+
+    PrinterEndo(
+      _.add(s"val ${service.descriptorName}: $grpcServiceDescriptor =")
+        .indent
+        .add(s"""$grpcServiceDescriptor.newBuilder("${service.getFullName}")""")
+        .indent
+        .addStringMargin(schemaDescriptor)
```
thesamet: Given the previous comment, this may be just a simple `add()`. If not implement using `call()`

thesamet:
Thanks for sending the PR. Can you explain in the PR description what wasn't functioning properly (impact of the problem).

jyane:
In my environment, ServerReflectionService does not work with ScalaPB.

Here is an example of reflection service (It does not work).
https://github.com/jyane/scalapb-example/blob/master/src/main/scala/jp/jyane/hello/GrpcMain.scala#L12-L16

I guess this PR will fix it?

dpratt:
@jyane You've got it exactly right - the ProtoReflectionService doesn't work if the BindableService doesn't have a proper ServiceDescriptor attached to it.

dpratt:
@thesamet I've pushed a commit that integrates your feedback.

PR #308, opened 2017-07-18
Update async observer completion.

dpratt:
- Handle StatusRuntimeExceptions as well as StatusExceptions on Observer
completion.
- Use tryFailure and trySuccess for Guava -> Scala Future translation.
This avoids a spurious exception thrown on (presumedly very rare)
multiple completion of the Future.


thesamet:
> Use tryFailure and trySuccess for Guava -> Scala Future translation.
This avoids a spurious exception thrown on (presumedly very rare)
multiple completion of the Future.

Aren't those spurious exceptions indicate a programming error (when a future is completed twice) and by switching to `trySuccess` we are silently hiding them?

dpratt:
That's a fair point, I've always used (and seen used) the tryComplete/tryFailure variants, but I'm happy to back that out.

dpratt:
Changes pushed.

PR #309, opened 2017-07-19
grpc-java 1.5.0

xuwei-k:


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -167,11 +167,12 @@ final class GrpcServicePrinter(service: ServiceDescriptor, override val params:
 
     p.addStringMargin(
       s"""val ${method.descriptorName}: $grpcMethodDescriptor[${method.scalaIn}, ${method.scalaOut}] =
-          |  $grpcMethodDescriptor.create(
```
xuwei-k: https://github.com/grpc/grpc-java/commit/17af756df42cda5d6dc1e5b4bca48ab2df053c21

PR #311, opened 2017-07-23
remove unused JavaConverters imports in FileDescriptor object file

xuwei-k:


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1182,7 +1182,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
       .call(generateMessageCompanion(message))
   }
 
-  def scalaFileHeader(file: FileDescriptor): FunctionalPrinter = {
+  def scalaFileHeader(file: FileDescriptor, javaConverterImport: Boolean = true): FunctionalPrinter = {
```
thesamet: can you make this not have a default value and change all callers to either pass file.javaConversions or false.

compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1194,7 +1194,7 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
          |
          |${if (file.scalaPackageName.nonEmpty) ("package " + file.scalaPackageName) else ""}
          |
-         |${if (file.javaConversions) "import scala.collection.JavaConverters._" else ""}""")
+         |${if (file.javaConversions && javaConverterImport) "import scala.collection.JavaConverters._" else ""}""")
```
thesamet: this would become `if  (javaConverterImport)`

xuwei-k:
done

PR #324, opened 2017-08-21
Add custom annotations for message

lkmtue:
This pull request makes it possible to add custom annotations to compiled messages.

Example:

```
message FooMessage {
  option (scalapb.message).annotations = "@CustomAnnotation";
  ...
}
```
will compile to
```
...
@SerialVersionUID(0L)
@CustomAnnotation
final case class FooMessage(
...
```


compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1316,6 +1319,10 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
 
     serviceFiles ++ enumFiles ++ messageFiles :+ fileDescriptorObjectFile
   }
+
+  private def getAnnotationList(message: Descriptor) = {
+    message.getOptions.getExtension[MessageOptions](Scalapb.message).getAnnotationsList().asScala.toSeq
```
thesamet: For consistency with all other options: please move this function to `MessageDescriptorPimps` in `DescriptorPimps.scala`. Rename it to `annotationsOption` and use the `scalaOptions` member to get the extension.

protobuf/scalapb/scalapb.proto
```diff
@@ -61,6 +61,8 @@ message MessageOptions {
 
   // Additional classes and traits to mix in to the companion object.
   repeated string companion_extends = 2;
+
+  repeated string annotations = 3;
```
thesamet: Add `// Custom annotations to add to the generated case class.`

e2e/src/main/protobuf/custom_annotation.proto
```diff
@@ -0,0 +1,13 @@
...
+package com.trueaccord.proto.e2e;
+
+import "scalapb/scalapb.proto";
+
+option (scalapb.options) = {
+  import: "com.trueaccord.pb.CustomAnnotation"
+};
+
+message CustomAnnotationMessage {
+  option (scalapb.message).annotations = "@CustomAnnotation";
```
thesamet: Please add another test with 3 different annotations, to verify that the generated code works in that case too.

lkmtue:
Thanks @thesamet for your review. Your comments have been resolved.

PR #337, opened 2017-09-12
Update scalapb.proto to use proto3 syntax

mvdstam:


thesamet:
This change may break people's code since the primitives won't be inside an option. What is the motivation for this change?

mvdstam:
Hi @thesamet, you're absolutely right. I was hoping the change from proto2 to proto3 would be more backwards-compatible.

The motivation for this change is that I'm trying to use protocol buffers in an environment where there is also PHP code present in some applications where there is Scala code present in others, where I can simply use the protobuf messages as a dependency for all of those applications.

Unfortunately, the [php support](https://github.com/google/protobuf/tree/master/php) for Protobuf seems to be severly lacking at the moment as proto2 and most well-known types aren't even working at the moment. I thought this change would be low-hanging fruit (and: fully backwards-compatible), but I was a bit too optimistic it would seem!

Closing the PR. Thanks for your quick response. 😃

thesamet:
One workaround could be to use the proto3 version you have only when generating PHP, the messages will stay compatible

PR #365, opened 2017-11-15
fix #364

### LINKED ISSUE #364: Code generation does not handle escaped strings properly
I'm getting that this line doesn't doesn't seem to generate the right scala code, because it seems to generate unescaped strings.
Using this sbt plugin
https://github.com/thesamet/sbt-protoc, with ScalaPB version 0.6.6
and having trouble with this line:
https://github.com/golang/protobuf/blob/master/proto/testdata/test.proto#L140
because it is generating this:
```scala
def getFStringDefaulted: String = fStringDefaulted.getOrElse("hello, "world!"\n") 
```
  
Where the problem is with the second `"`, which closes the String.

xuwei-k:


thesamet:
Thanks @xuwei-k

joshlemer:
Thankyou @xuwei-k for the very fast fix 🥇

PR #383, opened 2018-01-04
Add an option not to wrap some values in Options when generating Scala values

fabiangebert:
based on the changes by @eiennohito, I've created this pull request, including a test case.


compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
```diff
@@ -107,11 +107,11 @@ trait DescriptorPimps {
     // Is this field boxed inside an Option in Scala. Equivalent, does the Java API
     // support hasX methods for this field.
     def supportsPresence: Boolean =
-      fd.isOptional && !fd.isInOneof && (!fd.getFile.isProto3 || fd.isMessage)
+      fd.isOptional && !fd.isInOneof && (!fd.getFile.isProto3 || fd.isMessage) && !fieldOptions.getNoBox
 
     // Is the Scala representation of this field is a singular type.
     def isSingular = fd.isRequired || (
-      fd.getFile.isProto3 && !fd.isInOneof && fd.isOptional && !fd.isMessage)
+      fd.getFile.isProto3 && !fd.isInOneof && fd.isOptional && !fd.isMessage) || fieldOptions.getNoBox
```
thesamet: please change to (`|| (fieldOptions.getNoBox && fd.isOptional))

thesamet:
Please add a check to compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala that verifies that no_box was given on an optional field

e2e/src/test/scala/NoBoxSpec.scala
```diff
@@ -0,0 +1,19 @@
+import org.scalatest._
+import com.trueaccord.proto.e2e.no_box.Car
+import com.trueaccord.proto.e2e.no_box.Tyre
+import org.scalatest._
+
+class NoBoxSpec extends FlatSpec with MustMatchers {
+  val car = Car(tyre1 = Tyre(size = 10), tyre2 = Some(Tyre(size = 20)))
```
thesamet: Two additional things that should be tested to future-proof this code:
1. Say a remote system written in Java sends a message that doesn't specify this optional field. Can we parse it? What are we getting?

2. I think that were going to see some inconsistency with the Java version, that is worth documenting in a test:
Say we have Car(tyre1=Tyre.defaultInstance), does the field get serialized? We need to document (by test) that if we parse these bytes in Java, the parser would give `hasTyre1==false`. At the same time, if we use Java conversions on this message, on the Java side we will have `hasTyre1==true`
(I think there's no way around it, if we fixed the conversion to not set `tyre1` in Java, it will be lossy somewhere else)

fabiangebert: I think it's ok for a "no_box" field to say from a Scala perspective that missing field is equivalent to default instance.
I've added specs and pushed them a few hours ago.

PR #385, opened 2018-01-08
add deprecated annotation when [deprecated=true]

### LINKED ISSUE #377: add @scala.deprecated annotations when [deprecated=true]
> Fields with [deprecated=true] will be marked with @Deprecated in Java.

https://github.com/google/protobuf/blob/860d693cf70f13d823ab8e6f9118d94485fa95bd/CHANGES.txt#L1385

xuwei-k:
fix #377

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
@@ -155,7 +155,15 @@ trait DescriptorPimps {
 
     def fieldOptions: FieldOptions = fd.getOptions.getExtension[FieldOptions](Scalapb.field)
 
-    def annotationList = fieldOptions.getAnnotationsList().asScala.toSeq
+    def annotationList: Seq[String] = {
+      val deprecated = {
+        if (fd.getOptions.getDeprecated)
+          List(ProtobufGenerator.deprecatedAnnotation)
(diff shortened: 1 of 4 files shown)


compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
```diff
@@ -155,7 +155,15 @@ trait DescriptorPimps {
 
     def fieldOptions: FieldOptions = fd.getOptions.getExtension[FieldOptions](Scalapb.field)
 
-    def annotationList = fieldOptions.getAnnotationsList().asScala.toSeq
+    def annotationList: Seq[String] = {
+      val deprecated = {
+        if (fd.getOptions.getDeprecated)
+          List("""@scala.deprecated("", "")""")
```
thesamet: Please change to `deprecated(message="Marked as deprecated in proto file")`  for a friendlier message (presumably it would also make it easier to exclude by regex from various linters). The second parameter can be omitted since it defaults to `""`.

thesamet:
Thanks for working on this!

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
xuwei-k: done

PR #390, opened 2018-01-17
add equals to GeneratedEnum

ernisv:
Define equals method for GeneratedEnum. Previously case objects defaulted to object reference comparison, which do not work in some cases, f.e. when comparing case object with object deserialized using other serialization methods than Protobuf.

### DIFF
--- a/scalapb-runtime/shared/src/main/scala/scalapb/GeneratedMessageCompanion.scala
+++ b/scalapb-runtime/shared/src/main/scala/scalapb/GeneratedMessageCompanion.scala
@@ -25,6 +25,8 @@ trait GeneratedEnum extends Any with Product with Serializable {
     case _ => false
   }
 
+  override def hashCode = value
+
   def companion: GeneratedEnumCompanion[EnumType]
 
   def isUnrecognized: Boolean = false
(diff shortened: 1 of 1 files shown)


thesamet:
Wouldn't this make enums of different type equal as long as they have the same value?

ernisv:
I believe canEqual that is generated by case object should take care of
that using isInstanceOf.

thesamet:
Can you be more specific on the "other serialization methods" you are referring to above? Isn't it a bug in those serialization methods that they are giving you a second instance of what's supposed to be a singleton?

ernisv:
It was spotted using Kryo and I agree - in that particular case the issue is outside ScalaPB. But nevertheless - equality by value seems to be valid way to compare these enums.
I guess there may be some other scenarios where someone would compare objects for example loaded by different classloaders, etc. where object reference comparison would not work.

xuwei-k:
I think we should override `hashCode` as well if override `equals` in `GeneratedEnum`.

thesamet:
True that if this gets merged then we need to merge hashCode.

Overall, I am not so excited about this change as it is defensive code against another library that breaks our assumptions by cloning singletons. I'm willing to reconsider that if it's a huge problem for many people.

For the time being what are the possible workarounds?  For Kryo, you can write a custom serializer/deserializer that uses toByteArray/parseFrom (I've done it before, IIRC it was much more performant than the default Kryo behavior).

thesamet:
Closing for now. The problem that this PR is attempting to solve seems to be related to another serialization library that instantiates ScalaPB''s singletons.

PR #393, opened 2018-01-25
Change withX method to convert null to None

sjoerdmulder:
Current when using `.withX(null)` the generated code will create an object containing `Some(null)` which IMHO you want to prevent. Using `Option.apply` construct it will automaticly create either a None (in the case of null) or a Some(value)


thesamet:
Can you run `make_plugin_protos.sh` and add it as another commit so the tests will pass? Can you add a test to one of the e2e tests?

PR #408, opened 2018-02-22
fix name collision between oneof and ValueType

gregmflanagan:
This fixes a naming collision between `GeneratedEnumCompanion` and the generated code when `ValueType` or `value_type` is used as the oneof name.


thesamet:
Thanks for fixing this.

mrt181:
is this already released?
if not when will it be available?
I have the issue which is fixed by this.

thesamet:
I released 0.7.1 which contains this fix.

PR #419, opened 2018-03-22
Add `then` as reserved word for scala

tferi-da:
This solves errors like the following:
```
[error] <FILE>: then is a reserved word (since 2.10.0); usage as an identifier is deprecated
```


thesamet:
Thank you

tferi-da:
@thesamet when are you planning to publish the next release with this fix?

PR #429, opened 2018-04-11
Fix 427 and optionally tolerate forbidden field names

tferi-da:
scalaPB's current solution to forbidden field names assumes that the developer has control of all .proto sources, but that may not be the case. This PR
- Handles Unit message names
- pre/postFixes forbidden field names with an underscore according to the language.

Please review.


thesamet:
Thanks for working on this! I would prefer that we always allow the "forbidden names" by prefixing them with underscores rather than making them a user-controlled switch. We found that it is undesirable to have code that can only be compiled when specific compiler parameters are given (imagine that you need to compile code that requires conflicting values of a parameter). 

Finally, can you please run make_plugin_protos.sh to regenerate the generated code that is shipped with ScalaPB. This would make the CI tests pass.

thesamet:
Yes, looks like the current failures come from a different local protoc version. I'll review the PR soon.

thesamet:
Merged in 8ad541c4

tferi-da:
Tanks @thesamet, much appreciated

PR #431, opened 2018-04-12
#430 Optimize remaining cases where custom mappers are called twice during serialization

### LINKED ISSUE #430: Optimize remaining cases where custom mappers are called twice during serialization
This is a follow-up of #294

There are still a handful of cases left where the results of custom mappers are not cached within both `__computeSerializedValue()` and `writeTo`, which cause extra overhead during serialization.

Example of currently generated code from [custom_types.proto](https://github.com/scalapb/ScalaPB/blob/master/e2e/src/main/protobuf/custom_types.proto)

```
// optional
if (name.isDefined) { __size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_name.toBase(name.get).serializedSize) + com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_name.toBase(name.get).serializedSize }

// required
__size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_requiredName.toBase(requiredName).serializedSize) + com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_requiredName.toBase(requiredName).serializedSize

// collection
names.foreach(names => __size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_names.toBase(names).serializedSize) + com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_names.toBase(names).serializedSize)
```

Which should be optimized as:

```
// optional
if (name.isDefined) {
  val __value = com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_name.toBase(name.get)
  __size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(__value.serializedSize) + __value.serializedSize
}

// required
{
  val __value = com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_requiredName.toBase(requiredName)
  __size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(__value.serializedSize) + __value.serializedSize
}

// collection
names.foreach { __item =>
  val __value = com.trueaccord.proto.e2e.custom_types.CustomMessage._typemapper_names.toBase(__item)
  __size += 1 + _root_.com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag(__value.serializedSize) + __value.serializedSize
}
```

I'm going to send a PR shortly.

nktpro:
Fixes #430

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
@@ -602,38 +602,38 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
             } else {
               generateWriteSingleValue(field, toBaseType(field)(fieldNameSymbol))(printer)
             }
-          } else if (field.isSingular) {
-            // Singular that are not required are written only if they don't equal their default
-            // value.
+        } else if (field.isSingular) {
+          // Singular that are not required are written only if they don't equal their default
+          // value.
+          printer
+            .add(s"{")
+            .indent
+            .add(s"val __v = ${toBaseType(field)(fieldNameSymbol)}")
+            .add(s"if (__v != ${defaultValueForGet(field, uncustomized = true)}) {")
+            .indent
+            .call(generateWriteSingleValue(field, "__v"))
+            .outdent
+            .add("}")
+            .outdent
+            .add("};")
+        } else {
+          if (field.customSingleScalaTypeName.isDefined) {
             printer
-              .add(s"{")
+              .add(s"${fieldNameSymbol}.foreach { __v =>")
               .indent
-              .add(s"val __v = ${toBaseType(field)(fieldNameSymbol)}")
-              .add(s"if (__v != ${defaultValueForGet(field, uncustomized = true)}) {")
-              .indent
-              .call(generateWriteSingleValue(field, "__v"))
-              .outdent
-              .add("}")
+              .add(s"val __m = ${toBaseType(field)("__v")}")
+              .call(generateWriteSingleValue(field, "__m"))
               .outdent
               .add("};")
           } else {
-            if (field.customSingleScalaTypeName.isDefined) {
-              printer
-                .add(s"${fieldNameSymbol}.foreach { __v =>")
-                .indent
-                .add(s"val __m = ${toBaseType(field)("__v")}")
-                .call(generateWriteSingleValue(field, "__m"))
-                .outdent
-                .add("};")
-            } else {
-              printer
-                .add(s"${fieldNameSymbol}.foreach { __v =>")
-                .indent
-                .call(generateWriteSingleValue(field, toBaseType(field)("__v")))
-                .outdent
-                .add("};")
-            }
+            printer
+              .add(s"${fieldNameSymbol}.foreach { __v =>")
+              .indent
+              .call(generateWriteSingleValue(field, toBaseType(field)("__v")))
+              .outdent
+              .add("};")
           }
+        }
     }
       .when(message.preservesUnknownFields)(_.add("unknownFields.writeTo(_output__)"))
       .outdent
--- a/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
@@ -427,7 +427,16 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
     val fieldNameSymbol = fieldAccessorSymbol(field)
 
     if (field.isRequired) {
-      fp.add("__size += " + sizeExpressionForSingleField(field, toBaseType(field)(fieldNameSymbol)))
+      if (field.customSingleScalaTypeName.isDefined) {
+        fp.add(
+          "{",
+          s"  val __value = ${toBaseType(field)(fieldNameSymbol)}",
+          s"  __size += ${sizeExpressionForSingleField(field, "__value")}",
+          "};"
+        )
+      } else {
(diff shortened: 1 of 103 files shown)


compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -564,29 +591,49 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
...
-            .add("};")
-        } else {
-          printer
-            .add(s"${fieldNameSymbol}.foreach { __v =>")
-            .indent
-            .call(generateWriteSingleValue(field, toBaseType(field)("__v")))
-            .outdent
-            .add("};")
-        }
+            if (field.customSingleScalaTypeName.isDefined) {
```
nktpro: FYI there's an code indentation fix for the subsequent `else if` blocks here so the diff shows a little more code change than needed

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -443,14 +452,32 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
         fp.add(s"if (${toBaseType(field)(fieldNameSymbol)} != ${defaultValueForGet(field, true)}) { __size += ${sizeExpressionForSingleField(field, toBaseType(field)(fieldNameSymbol))} }")
       }
     } else if (field.isOptional) {
-      fp.add(s"if ($fieldNameSymbol.isDefined) { __size += ${sizeExpressionForSingleField(field, toBaseType(field)(fieldNameSymbol + ".get"))} }")
+      if (field.customSingleScalaTypeName.isDefined) {
```
thesamet: I might be missing something, but I don't see the difference between the two branches of the if. The first case is assigning `__value` as a temporary variable, the other one inlines the same computation. Why is this necessary?

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -443,14 +452,32 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
...
+        fp.add(
+          s"if ($fieldNameSymbol.isDefined) {",
+          s"  val __value = ${toBaseType(field)(fieldNameSymbol + ".get")}",
```
thesamet: Would it make sense to use this code for both cases (and then remove the `if`?) This would reduce duplication.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -443,14 +452,32 @@ class ProtobufGenerator(val params: GeneratorParams) extends DescriptorPimps {
...
     } else if (field.isRepeated) {
       val tagSize = CodedOutputStream.computeTagSize(field.getNumber)
       if (!field.isPacked) {
         Types.fixedSize(field.getType) match {
           case Some(size) => fp.add(s"__size += ${size + tagSize} * $fieldNameSymbol.size")
-          case None => fp.add(
-            s"$fieldNameSymbol.foreach($fieldNameSymbol => __size += ${sizeExpressionForSingleField(field, toBaseType(field)(fieldNameSymbol))})")
+          case None =>
+            if (field.customSingleScalaTypeName.isDefined) {
+              fp.add(
```
thesamet: Would it make sense to use this code for both cases (and then remove the `if`?) This would reduce duplication.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
thesamet: Would you be able to fix the indentation in a separate commit? It's difficult to see the difference.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
nktpro: I thought it through before I worked on it and decided to stay consistent with PR #333: basically trade-off code verbosity to avoid introducing any possible regression to the existing generated code for the more common cases where custom mappers are not used. 

Technically you're absolutely right there's no need for us to branch out for `field.customSingleScalaTypeName.isDefined` for any of these since the `if` block covers for both cases. I'm just not 100% sure whether the extra `val` assignment when it's only used once and there's no custom mappers involved would introduce any overhead. I'd assume JVM JIT is smart enough to make it irrelevant, just can't guarantee without actual measurement. Hence I chose the safer route.

Let me know if you still want to collapse both branches into just one to reduce code, I'll send in the commits.

thesamet: Let's collapse to a single code path. I'll be surprised if the JIT doesn't take care of this case.

nktpro:
All custom mapping code gen branches have been collapsed. As a consequence please note that the amount of changes in checked-in generated code will look pretty scary.

nktpro:
@thesamet if you feel this is now too invasive I can revert it back to branching like it was originally so it doesn't touch any existing code that doesn't make use of custom mapping and doesn't benefit from this change. Thoughts?

thesamet:
Merged - thank you. I'll be writing a few microbenchmarks in the future so it is easy to tell if there's a significant difference. I don't believe there's much of a difference in performance for this case.

PR #434, opened 2018-04-19
Migrations from Dotty

allanrenucci:
This changes were needed to make ScalaPB compile with Dotty (as part of our community build). I though I might as well upstream them

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorPimps.scala
@@ -433,12 +433,12 @@ trait DescriptorPimps {
     }
   }
 
-  implicit class EnumDescriptorPimp(val enum: EnumDescriptor) {
-    def parentMessage: Option[Descriptor] = Option(enum.getContainingType)
+  implicit class EnumDescriptorPimp(val `enum`: EnumDescriptor) {
+    def parentMessage: Option[Descriptor] = Option(`enum`.getContainingType)
 
-    def scalaOptions: EnumOptions = enum.getOptions.getExtension[EnumOptions](Scalapb.enumOptions)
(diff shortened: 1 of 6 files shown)


allanrenucci:
[Here](https://github.com/dotty-staging/ScalaPB/commit/02c1747cc6cecd906e29d47394d0d1a9c7cb6aa4) is what we use in our community build. The [dotty-example-project](https://github.com/lampepfl/dotty-example-project) has instructions on how to cross-compile with Scala 2.12 and Dotty. However, we are currently migrating the Dotty sbt plugin to sbt 1 and some of the documentation is outdated. Also, there is currently no stable version of Dotty for sbt 1, only nightly builds. A stable version with updated documentation should be released next week.

thesamet:
I tried today to cross-compile using sbt-dotty and made some progress. It would be nice if the project ScalaPB depends on (fastparse and scalatest) would publish a release for dotty.

allanrenucci:
In the future we plan to publish Dotty versions of the main Scala open source projects.  For now, you should be able to compile using the Scala 2.12 binaries.

Here is what we compile in our community build: https://github.com/dotty-staging/ScalaPB/blob/eb56a050babf28291a8f98bbda241802a82a0e21/build.sbt#L262-L266. Currently, compiling the tests doesn't work because of the dependency on scalatest and its use of macros.

thesamet:
Thanks, this is realy useful. I have implemented most of these migrations and some more on the dotty branch in this repo. I got the compiler plugin, lenses and the runtime environment to cleanly compile. The implicit conversions around lenses don't work well yet (https://github.com/lampepfl/dotty/issues/4647), which cause some errors in our e2e test. I need to keep investigating why. Also, it would be nice to continue this once utest, minitest or scalatest can work with dotty.

allanrenucci:
I am closing this PR then. We have a version of minitest working with Dotty at https://github.com/dotty-staging/minitest. We might publish it at some point

PR #454, opened 2018-06-06
Set scalaVersion and crossScalaVersions in native projects.

olafurpg:
Importing the build into IntelliJ failed by trying to resolve native
artifacts for Scala 2.12. This change sets the scalaVersion to 2.11 for
the native modules so that importing from IntelliJ works in a clean
cloned build.


thesamet:
Awesome - thank you @olafurpg

PR #458, opened 2018-06-13
Introduce new "sealed oneof" encoding for a subset of oneofs

### LINKED ISSUE #455: Proposal: sealed oneof for more ergonomic ADTs
Hi, I would like to be able to encode an sealed ADT using protobuf. 

```scala
case class Lit(value: Int) extends Expr
case class Add(lhs: Expr, rhs: Expr) extends Expr
sealed trait Expr
```

If I were to try and define the ADT above in protobuf I would do something like
```proto
message Lit {
  int32 value = 1
}
message Add {
  Expr lhs = 1;
  Expr rhs = 2;
}
message Expr {
  oneof value {
    Lit lit = 1;
    Add add = 2;
  }
}
```
However, the current oneof encoding in ScalaPB makes it difficult to construct expressions. For example, to create `Add(Lit(1), Add(Lit(2), Lit(3))` you need to write

```scala
Expr(
  Expr.Value.Add(
    Add(
      Some(Expr(Expr.Value.Lit(Lit(1)))),
      Some(
        Expr(
          Expr.Value.Add(
            Add(
              Some(Expr(Expr.Value.Lit(Lit(2)))),
              Some(Expr(Expr.Value.Lit(Lit(3))))
            )
          )
        )
      )
    )
  )
)
```

I would like to propose an alternative code generation technique for a special category of oneofs that meet the following criteria

- All fields of the oneof are messages defined in the same proto file
- The message containing the oneof contains a single oneof with no fields outside of the oneof

Such oneofs can be called "sealed oneof" and the generated code for sealed oneofs could look like this instead

```scala
case class Lit(value: Int) extends Expr with GeneratedMessage with Message[Lit] { ... }
case class Add(lhs: Expr, rhs: Expr) extends Expr with GeneratedMessage with Message[Lit] { ... }
sealed trait Expr {
  final def isEmpty: Boolean = this.isInstanceOf[Expr.Empty.type]
  def asExprMessage: ExprMessage
}
object Expr {
  case object Empty extends Expr with GeneratedMessage with Message[Expr] { ... }
}
final class ExprMessage(underlying: Expr with GeneratedMessage with Message[Expr], tag: Int)
    extends GeneratedMessage with Message[Expr] { ... }
```

I did a manual experiment with this encoding and validated that the types add up and it's possible to persist `Expr` to binary and read back. The full implementation can be seen in this gist https://gist.github.com/olafurpg/899bae7b17fbc5d991b4c0888c3ffc42

If there is interest, I am happy to work on this and submit a PR. I would love feedback on the design and hear your thoughts.

We are using ScalaPB in SemanticDB to encode the type system of Scala and the current protobuf encoding for `Type` is difficult to work with, https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#type

olafurpg:
This PR changes the code generation for a subset of oneofs that can be translated to idiomatic Scala sealed traits. Messages containing such oneofs are called "sealed oneof" and must meet the following criteria:

- all oneof fields are defined in the same file
- all oneof field types are distinct top-level message types
- the oneof name must be `sealed_value`
- the containing type of the oneof has no fields outside of the oneof

For example, given the following proto definition
```proto
message Left { string message = 1; }
message Right { int32 number = 1; }
message Either {
  oneof sealed_value {
    Left left = 1;
    Right right = 2;
  }
}
```
the generated code will look roughly like this
```scala
package a
case class Left(message: String) extends Either with scalapb.Message[Left]
case class Right(number: Int) extends Either with scalapb.Message[Right]
sealed trait Either {
  def toEitherMessage: EitherMessage
}
object Either {
  case object Empty extends Either
  val defaultInstance: Either = Empty
}
case class EitherMessage(sealedValue: EitherMessage.Value) extends scalapb.Message[EitherMessage] {
  def asEither: Either
  // ...
}
object EitherMessage {
  sealed trait Value
  object Value {
    case object Empty extends Value
    case class Left(value: a.Left) extends Value
    case class Right(value: a.Right extends Value
  // normal oneof encoding for `message Either` ...
}
```

Additionally, when fields have a sealed oneof type they automatically get the `(scalapb.field).no_box = true` option. This allow you to concisely construct values for recursive ADTs.

For example, given the ADT below

```proto
message Lit {
    int32 value = 1;
}

message Add {
    Expr lhs = 1;
    Expr rhs = 2;
}

message Expr {
    oneof sealed_value {
        Lit lit = 1;
        Add add = 2;
    }
}
```

You can construct a value like `Add(Lit(1), Add(Lit(2), Lit(3))`.

Fixes #455 .

TODOs:
- [ ] Decide on sealed oneof detection, should there be an option `(scalapb.message).sealed_oneof = true`;?
- [ ] Add documentation to scalapb website
- [ ] Add scalacheck generator for sealed oneofs, I would appreciate help here


olafurpg:
Mima complains about
```
[error]  * method sealedOneofs()scalapb.compiler.SealedOneofs in trait scalapb.compiler.DescriptorPimps is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scalapb.compiler.DescriptorPimps.sealedOneofs")
```
Is it OK to filter or should we find a way to work around this breaking change?

olafurpg:
An example of the generated code for SemanticDB.type can be seen here https://github.com/scalameta/scalameta/blob/74e7c070ae90363857b7abe576ed9f6369b2f6de/semanticdb/semanticdb3/src/main/generated/scala/meta/internal/semanticdb3/TypeMessage.scala

```proto
message Scope {
  repeated string symlinks = 1;
  repeated SymbolInformation hardlinks = 2;
}

message Type {
  enum Tag {
    // reserved 2, 3, 4, 5;
    UNKNOWN_TYPE = 0;
    TYPE_REF = 1;
    SINGLETON_TYPE = 15;
    INTERSECTION_TYPE = 16;
    UNION_TYPE = 17;
    WITH_TYPE = 18;
    STRUCTURAL_TYPE = 6;
    ANNOTATED_TYPE = 7;
    EXISTENTIAL_TYPE = 8;
    UNIVERSAL_TYPE = 9;
    CLASS_INFO_TYPE = 10;
    METHOD_TYPE = 11;
    BY_NAME_TYPE = 12;
    REPEATED_TYPE = 13;
    TYPE_TYPE = 14;
  }
  // reserved 3, 4, 5, 6;
  Tag tag = 1;
  TypeRef typeRef = 2;
  SingletonType singletonType = 16;
  IntersectionType intersectionType = 17;
  UnionType unionType = 18;
  WithType withType = 19;
  StructuralType structuralType = 7;
  AnnotatedType annotatedType = 8;
  ExistentialType existentialType = 9;
  UniversalType universalType = 10;
  ClassInfoType classInfoType = 11;
  MethodType methodType = 12;
  ByNameType byNameType = 13;
  RepeatedType repeatedType = 14;
  TypeType typeType = 15;
}

message TypeRef {
  Type prefix = 1;
  string symbol = 2;
  repeated Type type_arguments = 3;
}

message SingletonType {
  enum Tag {
    UNKNOWN_SINGLETON = 0;
    SYMBOL = 1;
    THIS = 2;
    SUPER = 3;
    UNIT = 4;
    BOOLEAN = 5;
    BYTE = 6;
    SHORT = 7;
    CHAR = 8;
    INT = 9;
    LONG = 10;
    FLOAT = 11;
    DOUBLE = 12;
    STRING = 13;
    NULL = 14;
  }
  Tag tag = 1;
  Type prefix = 2;
  string symbol = 3;
  int64 primitive = 4;
  string string = 5;
}

message IntersectionType {
  repeated Type types = 1;
}

message UnionType {
  repeated Type types = 1;
}

message WithType {
  repeated Type types = 1;
}

message StructuralType {
  reserved 1, 2, 3;
  Type tpe = 4;
  Scope declarations = 5;
}

message AnnotatedType {
  reserved 2;
  repeated Annotation annotations = 3;
  Type tpe = 1;
}

message ExistentialType {
  reserved 2;
  Type tpe = 1;
  Scope declarations = 3;
}

message UniversalType {
  reserved 1;
  Scope type_parameters = 3;
  Type tpe = 2;
}

message ClassInfoType {
  reserved 1, 3;
  Scope type_parameters = 4;
  repeated Type parents = 2;
  Scope declarations = 5;
}

message MethodType {
  reserved 1, 2;
  Scope type_parameters = 4;
  repeated Scope parameterLists = 5;
  Type return_type = 3;
}

message ByNameType {
  Type tpe = 1;
}

message RepeatedType {
  Type tpe = 1;
}

message TypeType {
  reserved 1;
  Scope type_parameters = 4;
  Type lower_bound = 2;
  Type upper_bound = 3;
}
```

It seems this PR needs to handle two more cases:

- support field with nested message types
- support two layers of sealed one_of, allow for example `SingletonType` to be a sealed oneof as well. This is necessary to encode ADTs like this
```scala
sealed trait Expr
case class Add(lhs: Expr, rhs: Expr) extends Expr
sealed trait Literal extends Expr // second layer of sealed one_of
object Lit {
  case class Double(value: scala.Double) extends Lit
  case class Int(value: scala.Int) extends Lit
}
```

thesamet:
Hey @olafurpg , thanks for this PR, it looks really good. I started to work on this locally and make the necessary changes to get it merged. Great work there, the changes are mostly around my personal preferences.

One thought in the meantime: what's the reason behind restricting the oneof cases to only top-level messages? Anything prevents them to be anywhere in the file?  I am thinking that (at least for the first version), we can restrict the oneof messages to be nested inside the sealed oneof message. This ensures they get generated within the companion object and not into separate files. This eliminates the need to have a special case where single_file is false, and the need for a sealed oneofs cache (since the search for a parent would be constant time). If we do this, then the format for sealed oneofs would then be limited to this:

```
message Either {
  message Left { string message = 1; }
  message Right { int32 number = 1; }

  oneof sealed_value {
    Left left = 1;
    Right right = 2;
  }
}
```

WDYT? Would that be too restrictive for your use case?

Another change I am making is that when there exists a oneof named `sealed_value`, code generation would abort if it doesn't pass the validation rules we defined (distinct message types, no other fields, etc)

olafurpg:
Thank you for the review, I appreciate you updating the changes to your preferences.

> One thought in the meantime: what's the reason behind restricting the oneof cases to only top-level messages? 

Good question, I added the limitation to keep the implementation simple but I don't see any fundamental reason why it should not work.

>  I am thinking that (at least for the first version), we can restrict the oneof messages to be nested inside the sealed oneof message.

I am fine with this restriction. I agree that it would be a nice simplification and avoid the hacky solution in this PR to print sub-messages in the sealed oneof file. In Scalameta, we can work around this limitation by introducing `val` and `type` forwarders in the package object to move the nested classes into the package scope.
```
val SingletonType = Type.SingletonType
type SingletonType = Type.SingletonType
val TypeRef = Type.TypeRef
type TypeRef = Type.TypeRef
...
```
We already have a `val NoType = Type.Empty` forwarder so this requirement would make things more consistent if anything.

> Another change I am making is that when there exists a oneof named sealed_value, code generation would abort if it doesn't pass the validation rules we defined (distinct message types, no other fields, etc)

Good idea 👍

thesamet:
Great! Will have the code finalized and merged soon.

olafurpg:
While testing out this new encoding @xeno-by came across that it would be nice to have an abstract ` def toProtoString: String` available on the sealed trait. What do you think about adding that?

thesamet:
Wouldn't `oneof.toBase.toProtoString` do the trick, or it's about making the access slightly shorter by bypassing the `toBase`?

My initial thinking is that if we add `toProtoString` to the sealed trait then it would be more consistent to have it extend the rest of the `GeneratedMessage` interface, but that brings us to the point you made [here](https://github.com/scalapb/ScalaPB/issues/455#issuecomment-395976346) about whether the call is viewed as applied to the case or the containing message.

olafurpg:
That is a good point, I agree it would give the impression that `(_: Supertype).toProtoString` should be different from `(_: Subtype).toProtoString` when that is not possible. Let's keep it explicit with `.toBase.toProtoString`

thesamet:
This has been merged via bef7ba91

Big thanks @olafurpg for making the proposal and implementing this.

Docs: https://scalapb.github.io/sealed-oneofs.html

Due to binary incompatibilities that have started to build up, this feature will come with ScalaPB 0.8 which will be shipping soon.

olafurpg:
Thank you @thesamet for cleaning up the implementation. I like the new changes, especially `asMessage` instead of `asExprMessage`.

Sealed oneofs have already been a big enabler for us in in SemanticDB, we're using it for the message types `Type`, `Signature` and `Constant`. https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#type

PR #464, opened 2018-06-22
execute scalafmtCheck in travis test

xuwei-k:


proptest.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env sh
 set -e
-sbt -J-Xmx4500M ++$TRAVIS_SCALA_VERSION test
+sbt -J-Xmx4500M ++$TRAVIS_SCALA_VERSION scalafmtCheck test:scalafmtCheck test
```
thesamet: The `__misc__` area in .travis.yml would be better IMO, since it runs one (and not for each Scala version)

xuwei-k: done

PR #468, opened 2018-06-26
Fully qualified scala.Predef.require in generated code

simuons:
If protobuf message contains field named `require` generated code fails to compile


thesamet:
Merged in bd4e2bdd - thanks for the PR

PR #471, opened 2018-07-06
move FieldMaskUtilPropSpec. fix compile error in js and native test

xuwei-k:
move `shared/src/test` => `jvm/src/test` because `FieldMaskUtilPropSpec` use java library (protobuf-java-util)


thesamet:
Thanks!

PR #473, opened 2018-07-06
fix loading project on Windows

codingismy11to7:


build.sbt
```diff
@@ -77,7 +77,7 @@ lazy val root =
       git.remoteRepo := "git@github.com:scalapb/scalapb.github.io.git",
       ghpagesBranch := "master",
       ghpagesNoJekyll := false,
-      includeFilter in ghpagesCleanSite := GlobFilter((ghpagesRepository.value / "api/scalapb/latest/*").getCanonicalPath)
+      includeFilter in ghpagesCleanSite := GlobFilter((ghpagesRepository.value / "api/scalapb/latest").getCanonicalPath + java.io.File.separator + "*")
```
thesamet: Thanks for looking into this. I am curious whether `ghpagesRepository.value / "api" / "scalapb" / "latest" / *` would work. Can you try?

codingismy11to7: I can try next week, but I can pretty much guarantee that it won't; in the end they wind up the same - a File with `*` as the `name`, and then the canonicalize on *nix jvm will accept it, and on windows jvm will puke

codingismy11to7: actually i took a few seconds to try it just now - same error

```
[info] Resolving key references (10057 settings) ...
[error] java.io.IOException: Invalid argument
[error]         at java.io.WinNTFileSystem.canonicalize0(Native Method)
[error]         at java.io.WinNTFileSystem.canonicalize(WinNTFileSystem.java:428)
[error]         at java.io.File.getCanonicalPath(File.java:618)
[error]         at $2e1bb8e4f5dabcf7145c$.$anonfun$root$9(build.sbt:80)
[error]         at scala.Function1.$anonfun$compose$1(Function1.scala:44)
[error]         at sbt.internal.util.EvaluateSettings$MixedNode.evaluate0(INode.scala:219)
```

PR #474, opened 2018-07-14
update protobuf version to 3.6

danbills:
I had a problem generating code from google's apis at github.com/googleapis/googleapis due to the use of `php_metadata_namespace`.  Turns out this was added in protobuf 3.6.

This update fixed that issue.


thesamet:
Thanks @danbills

PR #487, opened 2018-08-27
remove deprecated usage of addCallback

nadavwe:
Hi @thesamet,
we were trying to upgrade to guava 26.0 and then we found out that the flavour with no executor was deprecated and then removed on July 2018.
it's just a suggestion, feel free to edit/remove/solve it in any other way.
Thanks!
@simuons @vipo


thesamet:
Thanks!

Looks like this code should not change the existing behavior: https://github.com/google/guava/commit/87d87f5cac5a540d46a6382683722ead7b72d1b3#diff-3fe13f15fa4a5af9b4a55b21d7db2541L1275

PR #500, opened 2018-09-27
Fix #499 - Invalid code generated if file name contains reserved words

### LINKED ISSUE #499: Invalid Scala code generated if package names contain Scala reserved words
I have a protobuf that defines a package something like this:

```protobuf
package foo_override_bar_v1;
```
Which generates the following Scala code:
```scala
package foo.override.bar.v1
```
However, that package statement is invalid, because `override` is a Scala key word.

Could the code generator escape Scala keywords? For example:

```scala
package foo.`override`.bar.v1
```

blissd:


thesamet:
Can you also add a file to `e2e/src/main/protobuf` that demonstrates the problem this is fixing? The filename or its content should link to #499.

thesamet:
Thanks for reporting and fixing, @blissd - I'll have it released in the next few days.

Issue #507, opened 2018-10-04
Still reproduce MatchError when deserializing unknown enum value


syucream:
I've caught the MatchError in https://github.com/scalapb/ScalaPB/issues/285 in my environment with using ScalaPB.

And I could reproduce the error by below code.
https://github.com/syucream/scalapb-unknown-enum

```
$ date -u ; sbt "runMain Example" ; date -u
Thu Oct  4 09:29:20 UTC 2018
[info] Loading settings from plugins.sbt ...
[info] Loading project definition from /Users/syucream/sandbox/scalapb-unknown-enum/project
[info] Loading settings from build.sbt ...
[info] Set current project to scalapb-unknown-enum (in build file:/Users/syucream/sandbox/scalapb-unknown-enum/)
[info] Running Example
[error] (run-main-0) scala.MatchError: Some(scala.ref.WeakReferenceWithWrapper@1afc4206) (of class scala.Some)
[error] scala.MatchError: Some(scala.ref.WeakReferenceWithWrapper@1afc4206) (of class scala.Some)
[error]         at scalapb.descriptors.ConcurrentWeakReferenceMap.getOrElseUpdate(ConcurrentWeakReferenceMap.scala:13)
[error]         at scalapb.descriptors.EnumDescriptor.$anonfun$findValueByNumberCreatingIfUnknown$1(Descriptor.scala:128)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at scalapb.descriptors.EnumDescriptor.findValueByNumberCreatingIfUnknown(Descriptor.scala:125)
[error]         at scalapb.UnrecognizedEnum.scalaValueDescriptor(GeneratedMessageCompanion.scala:47)
[error]         at scalapb.UnrecognizedEnum.scalaValueDescriptor$(GeneratedMessageCompanion.scala:46)
[error]         at example.EnumExample$EnumField$Unrecognized.scalaValueDescriptor(EnumExample.scala:147)
[error]         at Example$.main(Main.scala:8)
[error]         at Example.main(Main.scala)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]         at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]         at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #508, opened 2018-10-04
#507 Get non WeakReference value when determining return value

syucream:
To fix https://github.com/scalapb/ScalaPB/issues/507 , I've tried using non WeakReference value.
ref. https://stackoverflow.com/questions/24109126/java-racing-against-the-garbage-collector


thesamet:
Thank you.

PR #517, opened 2018-11-21
Add support for async GRPC calls using cats.effect.IO

OlafurD:
This is an attempt to support GRPC calls using cats.effect.IO instead of scala futures. It creates a new stub implementation where the asyncUnaryCall is mapped from listenableFuture to IO instead of future.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -92,6 +92,7 @@ lazy val runtime = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     libraryDependencies ++= Seq(
       "com.lihaoyi" %%% "fastparse" % "1.0.0",
       "com.lihaoyi" %%% "utest" % "0.6.6" % "test",
+      "org.typelevel" %% "cats-effect" % "1.0.0",
       "commons-codec" % "commons-codec" % "1.11" % "test",
       "com.google.protobuf" % "protobuf-java-util" % protobufVersion % "test",
     ),
@@ -153,6 +154,8 @@ lazy val grpcRuntime = project.in(file("scalapb-runtime-grpc"))
     libraryDependencies ++= Seq(
       "io.grpc" % "grpc-stub" % grpcVersion,
       "io.grpc" % "grpc-protobuf" % grpcVersion,
+      "org.typelevel" %% "cats-core" % "1.4.0",
+      "org.typelevel" %% "cats-effect" % "1.0.0",
       "org.scalatest" %% "scalatest" % "3.0.5" % "test",
       "org.mockito" % "mockito-core" % "2.23.0" % "test"
     ),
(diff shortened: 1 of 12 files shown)


thesamet:
@OlafurD - Thanks for working on this! I haven't looked too deep but I was wondering how this compares with https://github.com/fiadliel/fs2-grpc? In general, I am open to having this feature merged. We would need to figure out a way to make this feature opt-in so users won't have to depend on cats-effects if they don't need to.

OlafurD:
I'm not familiar with fs2-grpc, but from a quick look it seems to be using fs2 stream to handle all grpc scenarios (unary, streaming, bidirectional). 

Do you have any suggestions on how to make this feature opt-in?

thesamet:
@OlafurD  sorry for the delay. To make this opt-in it would need to be a parameter a user passes to ScalaPB (like `flat_package`) to get this variant of the code generated. How would you address the streaming scenarios with this approach?

limansky:
Hi all. Shouldn't this be a separate library, to avoid unnecessary dependencies? For example, we are using ScalaPB in client library, and currently it's about 12 MB in a fat jar (we don't use GRPC, so we are not affected by this). I suppose adding cats + cats-effects might add about 20-30mb extra dependencies, that could be significant for the client side.

thesamet:
@limansky yes, if we proceed with this, it will be a separate runtime artifact so users who don't need it avoid the dependency on cats.

fiadliel:
> I'm not familiar with fs2-grpc, but from a quick look it seems to be using fs2 stream to handle all grpc scenarios (unary, streaming, bidirectional).

Just to comment on fs2-grpc functionality (2 months later 😛), the signatures for different modes are:

- unary to unary: `Input => F[Output]`
- unary to stream: `Input => Stream[F, Output]`
- stream to unary: `Stream[F, Input] => F[Output]`
- stream to stream: `Stream[F, Input] => Stream[F, Output]`

I would recommend this (or similar) set of signatures for any implementation using cats effect, observers are awkward if you're using the cats-effect ecosystem (including http4s which also uses FS2 for streaming). Also, I think it's probably preferable to use the underlying grpc-core API layer, rather than using the stub layer w/guava (guava adds an extra layer when you can use cats-effect concurrency primitives instead).

thesamet:
It seems that all the functionality discussed in this PR is already available in fs2-grpc. Can anyone comment on what would be a significant technical difference or benefit in having an alternative provided by ScalaPB? If there's none, I'm inclined to close this PR.

thesamet:
It seems that this PR isn't actively being worked on. Also, I don't believe we clarified the benefits over what is provided by fs2-grpc. I am closing for now, and happy to discuss further over our google group, gitter or a new issue.

PR #526, opened 2019-01-03
implement proto2 sealed trait without Empty as default value, use Option for optional sealed trait fields

shmishleniy:
For case of proto2 file

proto:
```
message PageType {
  oneof sealed_value {
    SimplePage t1 = 1;
    RedirectPage t2 = 2;
  }
}

message SimplePage {}
message RedirectPage {
  required string url = 1;
}
```

in old implementation you should always match impossible `Empty` case object in sealed trait however it useless because parseFrom will throw an exception on missing field.
proto:
```
message Page {
  required string id = 1;
  required PageType tpe = 2;
}
```
scala:
```
case class Page(id: String, tpe: PageType)
sealed trait PageType
sealed case object Empty extends PageType
case class SimplePage() extends PageType
case class RedirectPage(url: String) extends PageType
```
it can be `Empty` only if you set it with lense/copy

in this is implementation we can generate sealed trait without `Empty` case object in case of required field.
proto:
```
message Page {
  required string id = 1;
  required PageType tpe = 2;
}
```
scala:
```
case class Page(id: String, tpe: PageType)
sealed trait PageType
case class SimplePage() extends PageType
case class RedirectPage(url: String) extends PageType
```

and Option[PageType] field in case of optional where PageType sealed trait without `Empty` case object 
proto:
```
message Page {
  required string id = 1;
  optional PageType tpe = 2;
}
```
scala:
```
case class Page(id: String, tpe: Option[PageType])
sealed trait PageType
case class SimplePage() extends PageType
case class RedirectPage(url: String) extends PageType
```

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
@@ -331,7 +331,7 @@ class ProtobufGenerator(
            else
              ".set") + field.upperJavaName + (if (field.isEnum && field.getFile.isProto3) "Value"
                                               else "")
-        val scalaGetter = 
+        val scalaGetter =
           if (field.isSealedOneofType && field.isOptional && !field.isInOneof && field.getFile.isProto2 && !message.isMapEntry) {
             scalaObject + "." + fieldAccessorSymbol(field) + ".orNull"
           } else {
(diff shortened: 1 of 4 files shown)


thesamet:
Hi @shmishleniy , thanks for submitting this PR!

> in old implementation you should always match impossible Empty case object in sealed trait however it useless because parseFrom will throw an exception on missing field.

That's not true. A `oneof` may have at most one of its field set, but it is also valid for it to have none of its field set. An instance of `PageType` where none of the fields is set would correspond to `PageType.Empty` regardless of whether you are using optional or required fields to refer to this `PageType`. This would make a valid message and no exception would be thrown at parsing time.

The main problem with this PR as it is now, is that it would prevent users from encoding and decoding valid messages, those that an empty `PageType` instance assigned to their `tpe` field.

The other problem is proto2 and proto3 consistency - a sealed oneof may be defined in proto2 and be used in proto3 or the other way around. I think that the current code assumes the usage and definition of a sealed oneofs are always using the same proto language version. Ideally sealed ones work consistently regardless of the proto version.

shmishleniy:
Hi, 
Yes you are right I missed this case. You can serialize/deserialize with Empty if you set it so, but what is the reason to set required field to Empty and why this Empty value exists for required field. So basically we have no required sealed trait field.

Also in proto2 we have Option for primitives and messages, so sealed trait should be wrapped to Option too.
For proto3 we have no Option for primitives and have default value for that fields so we can have Empty for sealed trait as default value either.

Added few test on compatibility with proto3: 1e7f5f0

for optional field in proto2:
```
proto2 -> proto3 | None -> Empty
proto3 -> proto2 | Empty -> None
```
for required field in proto2: 
`proto3 -> proto2 | Empty -> Exception about missing field`

for missing field in proto2: 
`proto2 -> proto3 | no field -> Empty`

Seems this is logically right behavior. We discuss only `oneof sealed_value` case, all other cases will work as is.

thesamet:
I'd like to avoid the proposed solution for the time being. It is increasing the generator complexity by adding new special cases, and increases confusion by generating different types by the proto language version. The gain, on the other side, is too small in my view.

I'd be more open to a solution which is uniform across all proto versions.

@olafurpg I'm interested in your take on how the `Empty` case of sealed oneofs turned out to be. Would it be better if we:
(a) removed it and use `Option[SealedOneof]` all over instead? OR
(b) have the users manually specify somehow the `Empty` case to their sealed oneofs in they want them? In that world, if you parse a sealed oneof that doesn't support `Empty` and it comes in `Empty` the parser would throw.

olafurpg:
@thesamet the `Empty` sealed oneof case has worked out great in our experience from Scalameta. Using `Option[SealedOneof]` would be a regression since it makes both construction and pattern matching more complicated, even for simple cases

```diff
- Add(Lit(1), Lit(2))
+ Add(Some(Lit(1)), Some(Lit(2))
```

I would prefer exceptions over `Option` if I had to choose. iiuc, I like your (b) proposal where I could choose between having an explicit `Empty` or throwing an exception.

thesamet:
I'm closing the PR for now. Looks like the path forward with this should be to discuss over a github issue or Google group the desired behavior of sealed oneofs (ideally in a non breaking way).

PR #547, opened 2019-02-13
Fix generated code when package starts with build

ianoc-stripe:
The build method name is included in generated code which collides when used with some of the google api's under the build package space.

Added a failing test case, and included build as a special case with the context fields that we should include `_root_` for with


ianoc-stripe:
Would anyone have the cycles to give this a review? its pretty self-contained hopefully

thesamet:
Thanks for the PR! It turned out that it is sufficient to introduce the conflict check in a single place. I put the test case you provided under the `nojava` dir, to avoid adding `JavaThenScala` to the main e2e suite.

ianoc-stripe:
Excellent, thank you. Appreciate you taking the time to do it so it works well in the repo

PR #555, opened 2019-03-13
Lens optional set

codingismy11to7:
Give lenses a new method that takes an Option of the field type, and only set the field if the Option is nonEmpty

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -298,7 +298,13 @@ lazy val lenses = crossProject(JSPlatform, JVMPlatform, NativePlatform).in(file(
           )
       }
     },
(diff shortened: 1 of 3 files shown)


thesamet:
Thanks! Can you add unit tests for the two cases?

codingismy11to7:
wish i'd seen this earlier...about to go on vacation. told slack to remind me on the 25th to do it.

codingismy11to7:
and should i add exclusions for mima?

thesamet:
Yes please, that's fairly easy (check history of build.sbt for examples).

Issue #562, opened 2019-03-26
sealed oneof to allow for scalapb.message extends


aalkilani:
When using [Sealed oneofs](https://scalapb.github.io/sealed-oneofs.html#sealed-oneofs), I would like to have the generated sealed trait, or generated object extend a base trait.
Today (scalapb 0.8.4), using either
`option (scalapb.message).extends`
OR
`option (scalapb.message).companion_extends`
on the sealed oneof message definition has the effect that the generated Message case classes that extend from scalapb.GeneratedMessage get the additional mixin and not the sealed oneof trait.

Here's a practical example:
proto file:
```
message PlayerShim {
    option (scalapb.message).extends = "MyBaseTrait";

    oneof sealed_value {
        BasketBallPlayer BasketBallPlayer = 1;
        SoccerPlayer SoccerPlayer = 2;
    }
}

message BasketBallPlayer {
    required int32 something = 1;
}

message SoccerPlayer {
    required int32 something_else = 1;
}
```

The generated code produces
```
sealed trait PlayerShim {
  isEmpty
  isDefined
  asMessage
}
```
Notice how the PlayerShim sealed trait does not extend `MyBaseTrait` even though it's "instructed" to do so.

Similarly, the PlayerShimMessage, BasketBallPlayer, and SoccerPlayer case classes end up extending `MyBaseTrait` instead.

The ask here is to provide an option, perhaps `(scalapb.message).sealedoneof_extends` that allows trait mix-ins for the generated sealed trait.

One practical use case for this is serialization with AKKA. To serialize a sealed one of, it's necessary to get the message via `asMessage`. This works fine if you only had one sealed oneof to deal with. When you have a few, there's no "base trait" to rely on such that Scala type matching would work with and `asMessage` could subsequently be called. In short, if I had multiple sealed one of traits, I would have to account for each and every one of them instead of doing something like this:

case oneof: MyBaseTrait => oneof.asMessage

Ideally, even if an option like this `(scalapb.message).sealedoneof_extends`  doesn't exist, ideally, all generated sealed oneof traits would extend from a base trait that we can match on globally. Currently, the generated sealed oneof trait does not mixin any global `type` or trait.

thesamet:
Hi @aalkilani ,

I am fine with either the suggestions you made, and also with implementing both:
1. Add `sealed_oneof_extends` (repeated string). This would require a check in ProtoValidation that the message it is applied to is indeed a sealed oneof.
[...]

PR #564, opened 2019-04-02
implements sealed_one_of extensions and base classes re: #562

aalkilani:
This implements the issues discussed in #562 
Specifically this implements a scalapb message option called `sealed_oneof_extends` that allows users to mixin traits that the generated `Sealed oneof` trait extends from.
In addition, the generated `Sealed oneof` trait now also, by default, extends from and implements a base trait called `GeneratedSealedOneOf`.

This helps users implement code as such:
`case value : scalapb.GeneratedSealedOneOf => value.asMessage`

In the specific case of using `ScalaPBJsonSupport` from [akka-http-scalapb](https://github.com/rustedbones/akka-http-scalapb) something like this is now possible for `Sealed oneofs`:

```
(someActor ? SomeCommand).onComplete {
  case Success(value : scalapb.GeneratedSealedOneOf) => complete(StatusCodes.OK, value.asMessage)
  case .....
}
```


scalapb-runtime/shared/src/main/scala/scalapb/GeneratedMessageCompanion.scala
```diff
@@ -278,3 +278,10 @@ abstract class GeneratedFileObject {
   // Companion object of all top-level messages
   def messagesCompanions: Seq[GeneratedMessageCompanion[_]]
 }
+
+trait GeneratedSealedOneOf extends Any with Product with Serializable {
+  type MessageType = GeneratedMessage with Message[MessageType]
```
thesamet: Can we leave this abstract and leave this as an upperbound?

aalkilani: That was my initial intent however I don't know how that could be done without introducing type parameters to the trait which then defeats the purpose of having the trait altogether because you would basically end up having to specify the type when matching against it. So instead of being able to simply write something like this:

`case value : scalapb.GeneratedSealedOneOf => value.asMessage`

One would have to write

```
case value : scalapb.GeneratedSealedOneOf[SomeMessageType1] => value.asMessage
case value : scalapb.GeneratedSealedOneOf[SomeMessageType2] => value.asMessage
case value : scalapb.GeneratedSealedOneOf[SomeMessageType3] => value.asMessage
case value : scalapb.GeneratedSealedOneOf[SomeMessageType4] => value.asMessage
case value : scalapb.GeneratedSealedOneOf[SomeMessageType5] => value.asMessage
...
```
I may be mistaken, but I'm unaware of a method to to add an upper bound to a type without introduces type parameters or generics.  Thoughts?

thesamet: Wouldn't `type MessageType <: GeneratedMessage with Message[MessageType]` work?

aalkilani: I must've miss-typed something when I tried that and concluded it didn't work earlier. This should work. Thanks for the note. I'll have this updated tomorrow.

With that said, any chance we can get a new release soon with these changes included? I'd much rather rely on a full release vs a local build we reference in projects.

Thanks!

aalkilani:
Thanks for the help @thesamet . Latest commit should address your feedback w.r.t using an abstract type.

thesamet:
Sure, we can cut a release as soon as this change is merged.

aalkilani:
ProtoValidationSpec and e2e spec added. @thesamet please let me know if anything else is needed. Thanks

e2e/src/test/scala/SealedOneofSpec.scala
```diff
@@ -63,5 +63,12 @@ class SealedOneofSpec extends FlatSpec with MustMatchers {
         |}
       """.stripMargin)
   }
+
+  "message with sealed_oneof_extends" should "extend from custom user trait" in {
+    assertCompiles(
```
thesamet: Instead of `assertCompiles` it is sufficient to check if `PlayerShip.defaultInstance mustBe a [BaseTrait]` , also verify that the specific subtypes extend `BaseTrait`s. All these checks can happen at runtime rather than trigger an additional compile.

aalkilani:
@thesamet , all tests succeed now. Let me know if you see anything else that needs to be addressed but I believe this covers it.
Thanks again

thesamet:
Looking good, I'll merge and release today.

thesamet:
This has been merged via 82bbd699 and released as 0.9.0-M2. Thanks @aalkilani for this PR!

aalkilani:
That's awesome! Thanks @thesamet. Was there a plan to also push this to Maven similar to "https://mvnrepository.com/artifact/com.thesamet.scalapb/scalapb-runtime_2.12/0.9.0-M1" ?

Also, what's the process of updating the website's docs? Any plans on updating and do you need help doing that/mind if I submit a PR with documentation updates as well?

thesamet:
Hi there. This is already on mvn: 

http://central.maven.org/maven2/com/thesamet/scalapb/scalapb-runtime_2.12/0.9.0-M2/

What's shown on mvnrepository.com can lag behind by a few hours. The change I committed in the morning also included an update to the docs site: https://github.com/scalapb/ScalaPB/commit/82bbd69965934df77969c821d0c68f99260c32b8#diff-5f3afd95112b62835bb24bfa57f03256

This isn't reflected yet on scalapb.github.io due to build errors we currently have for Scala 2.13.0-M5 (which are unrelated to this change)

PR #569, opened 2019-04-10
Update dependencies

xuwei-k:


rlebran:
Last grpc version fix a memory leak due to the default grace time. We would like this to be merged as well if possible.

PR #602, opened 2019-06-19
Add scalapb import to custom types example

hilios:
Update the documentation to make explicit on Custom Type example the scalapb import.

I had a hard time debugging because I didn't know that this import was needed, and leading to weird exceptions.


thesamet:
Thanks for the PR

PR #625, opened 2019-07-29
Added schema descriptor to gRPC method

atty303:
Currently `ServerCall#getSchemaDescriptor` returns null.
Therefore, it is not possible to access custom option definitions.
Now `MethodDescriptor` provides `SchemaDescriptor`, i have access to custom options via extensions.
And I added the corresponding E2E test.

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
@@ -163,6 +163,7 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
       |    .setSampledToLocalTracing(true)
       |    .setRequestMarshaller(${marshaller(method.inputType)})
       |    .setResponseMarshaller(${marshaller(method.outputType)})
+      |    .setSchemaDescriptor(new _root_.scalapb.grpc.ConcreteProtoMethodDescriptorSupplier(${service.getFile.fileDescriptorObjectFullName}.javaDescriptor, "${service.getName}", "${method.getName}"))
       |    .build()
       |"""
     )
--- a/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
@@ -163,7 +163,7 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
       |    .setSampledToLocalTracing(true)
       |    .setRequestMarshaller(${marshaller(method.inputType)})
       |    .setResponseMarshaller(${marshaller(method.outputType)})
-      |    .setSchemaDescriptor(new _root_.scalapb.grpc.ConcreteProtoMethodDescriptorSupplier(${service.getFile.fileDescriptorObjectFullName}.javaDescriptor, "${service.getName}", "${method.getName}"))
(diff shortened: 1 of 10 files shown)


atty303:
Any comments?

scalapb-runtime-grpc/src/main/scala/scalapb/grpc/ConcreteProtoMethodDescriptorSupplier.scala
```diff
@@ -0,0 +1,19 @@
+package scalapb.grpc
+
+import com.google.protobuf.Descriptors
+import io.grpc.protobuf.{ProtoMethodDescriptorSupplier, ProtoServiceDescriptorSupplier}
+
+class ConcreteProtoMethodDescriptorSupplier(
+    fileDescriptor: Descriptors.FileDescriptor,
+    serviceName: String,
```
thesamet: instead of passing a string here and looking up by name below, let's pass the actual descriptors to the constructors (and return them directly through the getters below). See the various `javaDescriptorSource` methods in `DescriptorImplicits.scala`, please add such methods to `ServiceDescriptorPimp` and `MethodDescriptorPimp`

thesamet:
Thanks for the PR. One comment added.

atty303:
Thanks for a comment. It resolved in ba47753.

scalapb-runtime-grpc/src/main/scala/scalapb/grpc/ConcreteProtoMethodDescriptorSupplier.scala
```diff
@@ -0,0 +1,15 @@
+package scalapb.grpc
+
+import com.google.protobuf.Descriptors
+import io.grpc.protobuf.{ProtoMethodDescriptorSupplier, ProtoServiceDescriptorSupplier}
+
+class ConcreteProtoMethodDescriptorSupplier(
```
thesamet: It is sufficient to pass only `methodDescriptor`, since you can get service and file from a `MethodDescriptor`. See https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Descriptors.MethodDescriptor

compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -163,6 +163,7 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
       |    .setSampledToLocalTracing(true)
       |    .setRequestMarshaller(${marshaller(method.inputType)})
       |    .setResponseMarshaller(${marshaller(method.outputType)})
+      |    .setSchemaDescriptor(new _root_.scalapb.grpc.ConcreteProtoMethodDescriptorSupplier(${service.getFile.fileDescriptorObjectFullName}.javaDescriptor, ${service.javaDescriptorSource}, ${method.javaDescriptorSource}))
```
thesamet: Please change to `_root_.scalapb.grpc.ConcreteProtoMethodDescriptorSupplier.fromMethodDescriptor(...)` and add the appropriate method in the companion object that instantiates this class (using the MethodDescriptor alone, see other comment). The reason for not calling the constructor directly here is that over time signatures evolve and it's easier to achieve binary and source compatibility if signatures that may somewhat change have names.

atty303:
@thesamet I fixed the problem. How about that?

thesamet:
Merged via 9696ec7b. Thank you!

Issue #384, opened 2018-01-08
GeneratedCodeSpec fail when Scala 2.12.4 (success Scala 2.12.2)


xuwei-k:
```
Compiling Scala sources. /var/folders/hw/vqly1hs14p77yl70qbqzsxk00000gq/T/set_1672899296653351081/rMjHi/qg/kn/vclsi.scala:164: error: illegal cyclic reference involving class vclsi
object vclsi extends scalapb.GeneratedMessageCompanion[rMjHi.qg.kn.vclsi] with scalapb.JavaProtoSupport[rMjHi.qg.kn.vclsi, whQ7n.vclsi] {
                                                                                                                                 ^
/var/folders/hw/vqly1hs14p77yl70qbqzsxk00000gq/T/set_1672899296653351081/rMjHi/qg/kn/vclsi.scala:1460: error: illegal cyclic reference involving class vclsi
      object bkz1 extends scalapb.GeneratedMessageCompanion[rMjHi.qg.kn.vclsi.hjazh.u.bkz1] with scalapb.JavaProtoSupport[rMjHi.qg.kn.vclsi.hjazh.u.bkz1, whQ7n.vclsi.hjazh.u.bkz1] {
                                                                                                                                                                              ^
/var/folders/hw/vqly1hs14p77yl70qbqzsxk00000gq/T/set_1672899296653351081/rMjHi/qg/kn/vclsi.scala:167: error: value newBuilder is not a member of object whQ7n.vclsi
    val javaPbOut = whQ7n.vclsi.newBuilder
                                ^
/var/folders/hw/vqly1hs14p77yl70qbqzsxk00000gq/T/set_1672899296653351081/rMjHi/qg/kn/vclsi.scala:184: error: illegal cyclic reference involving class vclsi
    ru = if (javaPbSource.hasRu) Some(rMjHi.qg.kn.vclsi.hjazh.u.bkz1.fromJavaProto(javaPbSource.getRu)) else None,
                          ^
/var/folders/hw/vqly1hs14p77yl70qbqzsxk00000gq/T/set_1672899296653351081/rMjHi/qg/kn/vclsi.scala:516: error: illegal cyclic reference involving class vclsi
  object n extends scalapb.GeneratedMessageCompanion[rMjHi.qg.kn.vclsi.n] with scalapb.JavaProtoSupport[rMjHi.qg.kn.vclsi.n, whQ7n.vclsi.n] {
                                                                                                                                         ^
[...]

PR #627, opened 2019-07-31
Fix -Ybreak-cycles in ShemaGenerators.scala

lihaoyi-databricks:
Trying to run these tests locally fails on `++2.12.8` because the `-Ybreak-cycles` flag wasn't included (See https://github.com/scalapb/ScalaPB/issues/384). The interpolation was including the Seq wrapper in the interpolated string i.e. `-cp blah:blah:blah Seq(-Ybreak-cycles) -d blah`, which obviously doesn't work.

Not sure if this is the correct solution, but the previous code was clearly wrong


thesamet:
Shocking that this has been working... Thanks for fixing

PR #631, opened 2019-08-12
Add the file-level option `default_values_in_constructor`.

### LINKED ISSUE #629: Add an option for removing default parameter value in generated codes.
Currently, there is default value for optional field in codes generated by `scalapb`.
For example, there is the proto file `message foo { int32 x = 1; }` written in proto3.
Generated code by `scalapb` is like: `final case class Foo(x: _root_.scala.Int = 0)`.
The problem is that a programmer can omit putting `x` when encoding data to proto because there is no compile error due to the default value (`0`).
So, removing default value in generated codes by `scalapb` is good for preventing programmer's mistake.
So, I suggest adding new option for removing default parameter value.

taeguk:
Fix #629.


taeguk:
@thesamet I'm concerned about the name of the option. Another names what I thought are: `defaultValuesForFieldsInConstructor` and `defaultValuesInConstructor`.
Do you have any idea about name?

taeguk:
@thesamet Excuse me, ping?

thesamet:
Hi @taeguk , Thanks for sending the PR, I appreciate the effort. I responded over the main issue. Closing for now.

protobuf/scalapb/scalapb.proto
```diff
@@ -82,6 +82,9 @@ message ScalaPbOptions {
   // `scala.collection.immutable.Map` will be used.
   optional string map_type = 14;
 
+  // If true, default parameter values for fields in constructor will be generated.
+  optional bool default_parameter_values_for_fields = 15;
```
thesamet: please rename to `default_values_in_constructor = 15 [default=true]`

compiler-plugin/src/main/scala/scalapb/package.scala
```diff
@@ -19,19 +19,22 @@ package object scalapb {
...
-        "no_lenses"                   -> !options(Lenses),
-        "retain_source_code_info"     -> options(RetainSourceCodeInfo)
+        "flat_package"                           -> options(FlatPackage),
+        "java_conversions"                       -> options(JavaConversions),
+        "grpc"                                   -> options(Grpc),
+        "single_line_to_proto_string"            -> options(SingleLineToProtoString),
+        "ascii_format_to_string"                 -> options(AsciiFormatToString),
+        "no_lenses"                              -> !options(Lenses),
+        "retain_source_code_info"                -> options(RetainSourceCodeInfo),
+        "no_default_parameter_values_for_fields" -> !options(DefaultParameterValuesForFields)
```
thesamet: Let's keep it a file-level (package-level) parameter. Please remove this line.

taeguk:
@thesamet I applied your feedbacks.

taeguk:
This PR is actually merged through b8997f6246. So I close it.

PR #642, opened 2019-08-28
preserve oneof/regular field ordering in constructor

bjaglin:
Rationale: adding as last attribute a field generating a default value in the constructor parameters (repeated/optional) of the generated class should not require updating code instantiating this message without named parameters, no matter whether one of the existing field is a oneof (KO on master) or not (OK on master).

Unfortunately, achieving this means switching order of parameters to follow the declaration order, which is a breaking change.... 

```proto
message UpdatePublisherDirectSalesCreativeRequest {
  message FirstPartyMedia {
    required string asset_path = 1;
  }

  message ThirdPartyMedia {
    required string vast_url = 1;
  }

  message ThirdPartyTracking {
    required string tracking_event = 1;
    required string tracking_url = 2;
  }

  required int64 creative_id = 2;
  required string creative_name = 3;
  required int64 format_size_id = 4;
  required int64 format_creative_type_id = 5;

  oneof media {
    FirstPartyMedia first_party_media = 6;
    ThirdPartyMedia third_party_media = 7;
  }
  repeated ThirdPartyTracking third_party_trackings = 8; // new field being added
}
```

master:
```scala
final case class UpdatePublisherDirectSalesCreativeRequest(
    creativeId: _root_.scala.Long,
    creativeName: _root_.scala.Predef.String,
    formatSizeId: _root_.scala.Long,
    formatCreativeTypeId: _root_.scala.Long,
    thirdPartyTrackings: _root_.scala.Seq[tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.ThirdPartyTracking] = _root_.scala.Seq.empty,
    media: tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.Media = tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.Media.Empty
    )

// KO
UpdatePublisherDirectSalesCreativeRequest(
  0,
  "foo",
  0,
  Media.FirstPartyMedia(FirstPartyMedia("bar")
)
```

with this PR:
```scala
final case class UpdatePublisherDirectSalesCreativeRequest(
    creativeId: _root_.scala.Long,
    creativeName: _root_.scala.Predef.String,
    formatSizeId: _root_.scala.Long,
    formatCreativeTypeId: _root_.scala.Long,
    media: tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.Media = tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.Media.Empty,
    thirdPartyTrackings: _root_.scala.Seq[tv.teads.domains.campaign.transport.UpdatePublisherDirectSalesCreativeRequest.ThirdPartyTracking] = _root_.scala.Seq.empty
    )

// OK
UpdatePublisherDirectSalesCreativeRequest(
  0,
  "foo",
  0,
  Media.FirstPartyMedia(FirstPartyMedia("bar")
)
```


bjaglin:
Regarding backward compatibility: we could expose this new behavior behind an opt-out flag supported until the first/next major release to make the transition smoother.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -997,25 +1010,31 @@ class ProtobufGenerator(
...
-              s"__fieldsMap.get(__fields.get(${field.getIndex})).asInstanceOf[_root_.scala.Option[$typeName]]"
-            (transform(field) andThen FunctionApplication(field.oneOfTypeName)).apply(
-              e,
-              EnclosingType.ScalaOption
-            )
-          }
-          val expr =
-            elems.reduceLeft((acc, field) => s"$acc\n    .orElse[${oneOf.scalaTypeName}]($field)")
-          s"${oneOf.scalaName.asSymbol} = $expr\n    .getOrElse(${oneOf.empty})"
+            Seq(transform(field).apply(e, enclosingType = field.enclosingType))
```
thesamet: Why does the order here matter?

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -709,20 +712,28 @@ class ProtobufGenerator(
...
+            name = field.scalaName.asSymbol,
+            typeName = typeName,
+            default = ctorDefaultValue,
+            annotations = annotations(field)
+          )
         )
+
+      case oneOfField if oneOfField.isInOneof =>
+        val oneOf = oneOfField.getContainingOneof
+        if (!visitedOneOffsNames.contains(oneOf.getFullName)) {
```
thesamet: instead of this approach, maybe we can use `sortBy` to sort the list of `ConstructorField`s  (`regularFeilds ++ oneofFields`). We can have two different key functions and it will be picked based on the opt-out value. We probably need to add to `ConstructorField` the necessary info for the sorting functions (like field index, for one ofs it can be the index of first field)

thesamet:
@bjaglin I'm ok with putting the change behind an opt-out flag. Since it's breaking compatibility it would be introduced in the next major version and the opt-out flag will be removed later down the road. To facilitate this,

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
bjaglin: Are we https://github.com/scalapb/ScalaPB/blob/c3e44a0\/scalapb-runtime/non-jvm/src/main/scala/com/google/protobuf/wrappers/Int32Value.scala#L89? If I remember correctly, I updated that because of a compilation error when running e2e. But indeed, instead of cascading the order changes in the constructor, we could use named parameters for internal code invoking it though.

bjaglin: I had a closer look, and only the `oneofs` parameters were explicitely named. I am generalizing usage of named parameters for calling the constructor to avoid that.

bjaglin:
Amended the commit with the following changes:
- updated calls to the constructor to use named params everywhere to avoid depending on the order
- sorting is done as a second step, after generating the parameter list (instead of modifying the existing loop)
- the new, breaking behavior is applied by default (although that sounds weird in a patch release?!), but the legacy one can be brought back via:
  - `followFieldDeclarationOrder` in sbt settings
  - `no_follow_field_declaration_order` in `scalapbc`
  - `follow_field_declaration_order` as a file-level option (I am not sure whether it would be better at the message level, but I followed what was done for lenses)
- added e2e test to document/protect the legacy/new behavior
- documented the new flag (focusing on its breaking nature)

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 
 ## [0.9.1 (Unreleased)](https://github.com/scalapb/ScalaPB/tree/v0.9.0...HEAD)
+- Order of constructor parameters in generated classes for messages now matches
```
bjaglin: removing this for now since it won't go in the next patch anyway

thesamet:
To make this easier to review and merge later, can you split this commit into two commits: one with the generator changes, and the other with all the automated changes you get by running `generate_plugin_proto.sh`?

bjaglin:
Splitted - I left the `scalapb.proto`-specific changes in the first commit for the compilation not to fail

bjaglin:
Rebased against https://github.com/scalapb/ScalaPB/commit/b8997f62460e445b65e9385e58c2253d11701270

protobuf/scalapb/scalapb.proto
```diff
@@ -85,6 +85,10 @@ message ScalaPbOptions {
   // If true, no default values will be generated in message constructors.
   optional bool no_default_values_in_constructor = 15;
 
+  // If false, oneof fields appear last in the constructor parameters list for
+  // generated messages, no matter where they are declared in the proto file.
+  optional bool follow_field_declaration_order = 16 [default=true];
```
bjaglin: this is now a bit inconsistent with the freshly-introduced, "negative" `no_default_values_in_constructor`, but follows the style of `lenses` - I have a preference for avoiding negation in the flag name.

thesamet: Yeah, I am not a fan of the negation in the flag name either. However, when the default value is `true`, it gets the `no_` suffix when it becomes a generator parameter which makes it a little inconsistent, so I have preference that the default value is false. Maybe you change it to `oneofs_after_fields_in_constructor`? Then it's non-negated and defaults to false. Open to other naming suggestions that follow the guidelines. I am thinking to rename `no_default_values_in_constructor` to something like `require_values_in_constructor` before the release.

bjaglin:
proptests are failing on 2.13
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000007cc000000, 154140672, 0) failed; error='Not enough space' (errno=12)

#

# There is insufficient memory for the Java Runtime Environment to continue.

# Native memory allocation (mmap) failed to map 154140672 bytes for committing reserved memory.

# An error report file with more information is saved as:

# /home/travis/build/scalapb/ScalaPB/hs_err_pid5367.log

Exception in thread "Thread-72" java.io.EOFException

	at java.base/java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:3062)

	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1561)

	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:430)

	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:822)

	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:811)

	at java.base/java.lang.Thread.run(Thread.java:834)
````

thesamet:
@bjaglin this occasionally happens on travis, due to large footprint of the scala compiler (which is invoked to compile the generated code)... Retrying should help.

bjaglin:
OK, I'll see if I am more lucky after I amend the commits to follow your flag name suggestion (unless I find a better name - will think it through for a little while).

bjaglin:
Updated the flag name and rebased against the mima updates on master.

bjaglin:
Rebased against https://github.com/scalapb/ScalaPB/commit/0a2d7b8a5f53d8f81a220539cf69d707f515b1a3, which made me add an unrelated commit about formatting

bjaglin:
Rebased against https://github.com/scalapb/ScalaPB/commit/9217f102d3a03666a695a4c7cf3594598ebbf893, https://github.com/scalapb/ScalaPB/commit/97100dc4c11aa40ac9fa082e2353d23c81b04ddf & https://github.com/scalapb/ScalaPB/commit/7604680e07240718479b24e03fa1986d3a09bdf7

bjaglin:
> Since it's breaking compatibility it would be introduced in the next major version and the opt-out flag will be removed later down the road. Some comments added.

By next major version, I guess you mean minor (by semver standards) ? Just wanted to align expectations, since keeping this up to date against master lately has been demanding :)

thesamet:
Correct, I meant minor. I'm thinking of releasing a 1.0.0 very soon. No need for you to keep rebasing, I can take care of that when it's time to merge.

bjaglin:
@thesamet depending on your 1.0.0 plans, what would be a rough ETA for you to consider merging this - weeks? months?

thesamet:
@bjaglin few weeks

thesamet:
Thanks for submitting the PR. This gets in 0.10.x. I ended up stripping out the generator parameters to keep things simple.

Issue #633, opened 2019-08-13
(Sealed) OneOf without Empty case


sideeffffect:
Currently, for `one_of` (both `sealed_value` or normal `one_of`s) generates not just the normal described cases, but also an extra `Empty` value. E.g.

```proto
message ContainingMyOneof {
  oneof nonsealed_value {
    HelloRequest request = 1;
    HelloReply reply = 2;
  }
}
```

generates

```scala
final case class ContainingMyOneof(nonsealedValue: ….ContainingMyOneof.NonsealedValue)
…
sealed trait NonsealedValue …
case object Empty extends ….ContainingMyOneof.NonsealedValue
final case class Request(value: ….HelloRequest) extends ….ContainingMyOneof.NonsealedValue
final case class Reply(value: ….HelloReply) extends ….ContainingMyOneof.NonsealedValue
```

We would need ScapaPB not to generate these extra `Empty` case objects.
I see at lease two ways how to do that:

 1. `ContainingMyOneof#nonsealedValue` would not be `ContainingMyOneof.NonsealedValue`, but `Option[ContainingMyOneof.NonsealedValue]`. That would be the most regular approach, ScalaPB encodes other fields this way too. The generated code would look like:
```scala
final case class ContainingMyOneof(nonsealedValue: Option[….ContainingMyOneof.NonsealedValue])
…
sealed trait NonsealedValue …
final case class Request(value: ….HelloRequest) extends ….ContainingMyOneof.NonsealedValue
final case class Reply(value: ….HelloReply) extends ….ContainingMyOneof.NonsealedValue
```

 2. ScapaPB's decoder would throw an exception if it would encounter a would be `Empty` input. The generated code would look like:
```scala
final case class ContainingMyOneof(nonsealedValue: ….ContainingMyOneof.NonsealedValue)
…
sealed trait NonsealedValue …
final case class Request(value: ….HelloRequest) extends ….ContainingMyOneof.NonsealedValue
final case class Reply(value: ….HelloReply) extends ….ContainingMyOneof.NonsealedValue
```

Both of these solutions would be significant changes, so I presume they could be non-default, hidden behind some king of option/setting.

I would contribute the code to do that, but wanted to discuss this with ScalaPB maintainers before writing code.

relevant tickets/PRs:
https://github.com/scalapb/ScalaPB/issues/455
https://github.com/scalapb/ScalaPB/pull/526
https://github.com/scalapb/ScalaPB/issues/578

[...]

PR #652, opened 2019-09-12
Enum Recognized

sideeffffect:
analogous to https://github.com/scalapb/ScalaPB/commit/4da080105f4fc6a56715afcf631720bc6688d2f0 and https://github.com/scalapb/ScalaPB/issues/633


sideeffffect:
@thesamet could I ask you for an advice? I'm afraid I'm stuck :disappointed: 

depending on whether I
 * amend `scalapb-runtime/*` enums, I fail on [MIMA check for Scala 2.11](https://travis-ci.org/scalapb/ScalaPB/builds/584663143)
 * don't amend, I fail on [EXAMPLES_AND_FORMATTING](https://travis-ci.org/scalapb/ScalaPB/builds/584385971)

thesamet:
The mima errors are legit, since there are some new methods and possibly some signature changes. You need to add exclusion filters in `build.sbt`. You can test locally with `mima.sh`

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -79,6 +82,7 @@ class ProtobufGenerator(
       }
       .add(s"  case __other => Unrecognized(__other)")
       .add("}")
+      .add(s"sealed trait $RecognizedEnumType extends $name")
```
thesamet: can the declaration come before the various usages? this now appears after.

thesamet: do we get a compilation error when one of the enum values is `Recognized`? If so, we should mutate the name of the enum value to avoid conflict.

sideeffffect: As the newly added test proves, an enum values `Recognized` doesn't clash with the subtype.
That's because the enum value is a `case object`, while the subtype is a `trait`.

sideeffffect:
I'm now regularly failing with `Not enough space` in Scala 2.13.0 `proptest.sh`:
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000007d7000000, 92274688, 0) failed; error='Not enough space' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 92274688 bytes for committing reserved memory.
```

have you @thesamet ever come across these memory issues with ScalaPB before? would you suggest something?

sideeffffect:
so now it didn't fail on memory, but with
```
Compiling Scala sources. Exception in thread "Thread-76" java.io.EOFException
	at java.base/java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:3062)
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1561)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:430)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:822)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:811)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

and the failing protos are supposed to be [proto.tar.gz](https://github.com/scalapb/ScalaPB/files/3623476/proto.tar.gz)
but I don't really understand what's wrong with the protos or generated Scala/Java. Sorry to be a bother @thesamet, but do you think you could have a look at them to see what's wrong?

sideeffffect:
@thesamet we have a green build! can we merge it? :pray:

sideeffffect:
@thesamet gentle ping

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
thesamet: See comment here https://github.com/scalapb/ScalaPB/pull/652/files#r324930710, looks good otherwise.

thesamet:
See comment on moving the declaration of the `Recognized` trait.

sideeffffect:
@thesamet cool, I've updated it according to your comment. build is green, I think we're ready to merge :tada:

thesamet:
Merged - thank you!

PR #658, opened 2019-09-16
Fix case for enum type

khlumzeemee:
Fixed a few problems encountered when running the code.


thesamet:
Wow, this has been there for long time. Thank you!

PR #660, opened 2019-09-17
Fix typo

khlumzeemee:
I thought about adding the address message definition somewhere too, maybe to consider?

```
message Address {
  optional string street = 1;
  optional string city = 2;
}
```


thesamet:
Yes, adding `Address`  definition next to where `Person` is defined would be helpful. Thanks!

Issue #353, opened 2017-10-02
Strip common enum prefixes


jsw:
Has there been consideration to adding an option to strip common enum value prefixes such that

```
enum ImageSize {
  IMAGE_SIZE_SMALL = 0;
  IMAGE_SIZE_LARGE = 1;
}


enum ButtonSize {
  BUTTON_SIZE_SMALL = 0;
  BUTTON_SIZE_LARGE = 1;
}
```

would result in something like:

```
object ImageSize {
  case object SMALL extends ImageSize ...
  case object LARGE extends ImageSize ...
}


object ButtonSize {
  case object SMALL extends ButtonSize ...
  case object LARGE extends ButtonSize ...
}
```

If not, is there a better workaround for dealing with the enum scoping constraint?

thesamet:
PR will be welcome. Probably something like: `strip_prefix="IMAGE_SIZE_"`

jsw:
I think a global option that follows what the C# compiler does would work well

https://developers.google.com/protocol-buffers/docs/reference/csharp-generated#enum

jeff

thesamet:
Will not do for now to keep things simple.

PR #730, opened 2020-01-09
scalapb/ScalaPB#353 and more

eugeneatnezasa:
Introduced ~~two~~ one file option:
~~1. 'enum_value_name_as_scala'
If true, then scala name will be used as enum 'name' value. Original proto name will be available as 'protoName' enum value field, 'fromName' method will resolve both variants (scala name and proto name) into enum value~~

2. 'enum_strip_prefix' see for issue #353
Indicate if prefix (enum name + optional underscore) should be removed in scala name.

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
@@ -80,9 +80,7 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
 
   implicit class AsSymbolPimp(val s: String) {
     def asSymbol: String =
-      if (SCALA_RESERVED_WORDS.contains(s) ||
-          // starts with number
-          s(0) >= '0' && s(0) <= '9') s"`$s`"
+      if (SCALA_RESERVED_WORDS.contains(s) || s(0).isDigit) s"`$s`"
       else s
   }
 
@@ -792,7 +790,7 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
         val enumValueName: String =
           if (enumValue.getFile.scalaOptions.getEnumStripPrefix) {
             val enumName = enumValue.getType.getName
-            val prefixes = Seq(enumName + "_", enumName, NameUtils.toAllCaps(enumName) + "_", NameUtils.toAllCaps(enumName))
+            val prefixes = Seq(enumName, NameUtils.toAllCaps(enumName)).flatMap(n => Seq(n + "_", n))
             val commonPrefix = prefixes.find(enumValue.getName.startsWith(_)).getOrElse("")
             enumValue.getName.stripPrefix(commonPrefix)
           } else enumValue.getName
@@ -800,10 +798,6 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
           allCapsToCamelCase(enumValueName, true)
         else enumValueName
       }
-
-    def name: String =
-      if (enumValue.getFile.scalaOptions.getEnumValueNameAsScala) scalaName
-      else enumValue.getName
 
     def isName = {
       Helper.makeUniqueNames(
--- a/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
@@ -786,15 +786,6 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
     def valueExtends: Seq[String] =
       enumValue.getType.nameSymbol +: scalaOptions.getExtendsList.asScala.toSeq
 
-    private def names = {
-      val l = enumValue.getType.getValues.asScala.map {d => d -> (if (d.scalaOptions.hasScalaName) Some(d.scalaOptions.getScalaName) else None)}
-      if (l.exists(_._2.nonEmpty)) {
-        // we have at least one scala name set
-      } else {
-        l.map(_._1.getName)
-      }
-    }
-
     def scalaName: String =
       if (scalaOptions.hasScalaName) scalaOptions.getScalaName
       else {
@@ -838,15 +829,6 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
         .map(t => t.getLeadingComments + t.getTrailingComments)
         .map(Helper.escapeComment)
         .filter(_.nonEmpty)
-    }
-
-    private def getCommonPrefix(head: String, tail: Seq[String]): String = {
-      val prefix = head.zipWithIndex.takeWhile { case (char, index) =>
-        tail.forall(s => s(index) == char)
-      }.map(_._1).mkString
-
-      if (prefix.indexOf("_") > 0) prefix.reverse.dropWhile(_ != '_').reverse
-      else prefix
     }
   }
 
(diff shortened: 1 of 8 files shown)


thesamet:
Thanks for the PR @eugeneatnezasa , can you add more details about the specific situation that leads you to want file option (1)? File option (2) makes sense. I'm curious why not just change the enum's name at the proto-level to the scala name and then not needing that option.

eugeneatnezasa:
@thesamet 
i came to these two during my work trying to convert pretty rich inhouse model written in pure scala to proto-based. Main problem I have for now is that enum value names scoping is different from what we have in scala (i read about c/c++ scoping) and because of that I cannot implement similar model (e.g. `enum Supplier(CompanyNameX, CompanyY)` and `enum Provider(CompanyNameX, CompanyNameZ)`, or some common words used in many same level enums like Hidden, Disabled etc.).
I can use `scala_name` (even if this makes whole source a bit ugly), but we use enums also for serialisation in json and this is exactly the cause for first option: I need same names for different enum values and I need to serialize/deserialize them to/from string exactly with these names.

So, I can live just with first option implemented, but because I started my initial question mentioning this strip prefix issue (and since it makes proto code also nicer/unified) I implemented it as well.

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -79,7 +79,11 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
   private lazy val fileOptionsCache = FileOptionsCache.buildCache(files)
 
   implicit class AsSymbolPimp(val s: String) {
-    def asSymbol: String = if (SCALA_RESERVED_WORDS.contains(s)) s"`$s`" else s
+    def asSymbol: String =
+      if (SCALA_RESERVED_WORDS.contains(s) ||
+          // starts with number
+          s(0) >= '0' && s(0) <= '9') s"`$s`"
```
thesamet: Nit... `s(0).isDigit`, the comment above is unnecessary.

thesamet:
I prefer to avoid introducing the first option (`enum_value_name_as_scala`) to avoid the additional complexity (name, protoName) and the implied change in contracts that can impact other libraries (are we sure users would want the scala name used in JSON?). My recommendation would be to use prefixes as the [style guide suggests](https://developers.google.com/protocol-buffers/docs/style), and if you use scalapb-json4s you can register a custom enum formatter that can strip the prefix.

If possible, can you remove the first option from the implementation and we can focus on the 2nd one. Please also add a test to e2e that verifies that the prefix is stripped (and it works for both EnumValueNaming options)

eugeneatnezasa:
> I prefer to avoid introducing the first option (enum_value_name_as_scala) to avoid the additional complexity (name, protoName) and the implied change in contracts that can impact other libraries (are we sure users would want the scala name used in JSON?). 

Using scala name as `name` is optional from my implementation. Currently I can get scala name (if I want to use it for serialization) only via `getClass.getName` and dealing with removing $ and maybe prefix.. Which is a bit weird.. My option allows to simplify this if required..

> My recommendation would be to use prefixes as the style guide suggests, and if you use scalapb-json4s you can register a custom enum formatter that can strip the prefix.

well, i cannot simply use prefixes (if i understood you correctly). Unfortunately strigified jsons (with enums) are persisted, so changing enum value names would be a pain...

> If possible, can you remove the first option from the implementation and we can focus on the 2nd one. Please also add a test to e2e that verifies that the prefix is stripped (and it works for both EnumValueNaming options)

of cause... you are maintainer 🙂at the end i can have version extended with scala name in our inhouse repo...
will try to do changes during weekend.

thesamet:
Not sure my recommendation above was put clearly. In the proto, name the enum values with a prefix:
```
enum Supplier {
  SUPPLIER_COMPANY_X = 1;
  SUPPLIER_COMPANY_Y = 2;
}
```

In scalapb-json4s, you can register an [EnumFormatter](https://github.com/scalapb/scalapb-json4s/blob/cdb4d2b474952781bb6ed1eb345447cdde5bfbfa/src/main/scala/scalapb/json4s/JsonFormat.scala#L57-L67) that would strip the `SUPPLIER_` prefix and get you the name you want.

Another option you have would be to nest the enums under different messages (or different packages) so you can have the same value names in two enums:

```
message SupplierConstants {
  enum Supplier {
    SUPPLIER_COMPANY_X = 1;
    SUPPLIER_COMPANY_Y = 2;
  }
}
```

Note sure I followed the following comment:
> well, i cannot simply use prefixes (if i understood you correctly). Unfortunately strigified jsons (with enums) are persisted, so changing enum value names would be a pain...
In the JSON format the keys are strings and enum values are strings. You can configure scalapb-json4s to use numbers for enums to avoid the problem of values being renamed causing problems. But even keys can get renamed over time, therefore, if you persist - it is recommended to use the binary format and not the JSON format.

thesamet:
The custom enum formatter recommended above can use the scala name (available in the descriptor), so you don't need to resort to getClass.getName, stripping `$`s, etc.

eugeneatnezasa:
Ok, make sense, thank you!

eugeneatnezasa:
I removed first option, added tests and applied other requested changes.

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -784,9 +786,18 @@ class DescriptorImplicits(params: GeneratorParams, files: Seq[FileDescriptor]) {
...
     def scalaName: String =
       if (scalaOptions.hasScalaName) scalaOptions.getScalaName
-      else if (enumValue.getFile.scalaOptions.getEnumValueNaming == EnumValueNaming.CAMEL_CASE)
-        allCapsToCamelCase(enumValue.getName, true)
-      else enumValue.getName
+      else {
+        val enumValueName: String =
+          if (enumValue.getFile.scalaOptions.getEnumStripPrefix) {
+            val enumName = enumValue.getType.getName
+            val prefixes = Seq(enumName, NameUtils.toAllCaps(enumName)).flatMap(n => Seq(n + "_", n))
```
thesamet: I prefer to have a simpler rule. Let's keep oen prefix - `NameUtils.toAllCaps(enumName) + "_"` , and omit the other three options. The concern is unintended consequences when there's no intention to strip: `enum C { CAT = 1; }` is not necessarily expected to be stripped to just `AT`.

docs/src/main/markdown/customizations.md
```diff
@@ -78,6 +79,9 @@ enums to a single Scala file.
 
 - Typically, enum values appear in UPPER_CASE in proto files, and ScalaPB generates case objects
   with exactly the same name in Scala. If you would like ScalaPB to transform the names into CamelCase, set `enum_value_naming` to `CAMEL_CASE`.
+  
+- To avoid setting scala name for each enum value, you can instead set `enum_strip_prefix` to `true`. In this case prefix if it contains enum name plus optional underscore or upper-case enum nape with optional underscore for enum values will be stripped.
```
thesamet: - It is a common practice in protobufs to prefix each enum value name with the name of the enum. For example, an enum name `Size` may have values named `SIZE_SMALL` and `SIZE_LARGE`. When you set `enum_strip_prefix` to true, ScalaPB will strip the enum's name from each value name, and they would become `SMALL` and `LARGE`. Then the name can be transformed to camel-case according to `enum_value_naming`. Note that the prefix that is removed is the all-caps version of the enum name followed by an underscore.

thesamet:
Merging. I'll take it from here. Thank you!

PR #759, opened 2020-02-12
Fix broken links for custom options examples

nikhilpatil:
I believe these are the correct links to the examples in repo. Thanks for ScalaPB :)


thesamet:
Thank you

PR #805, opened 2020-04-15
Add the `asRecognized` to parent trait GeneratedEnum

sideeffffect:
and document the method in documentation

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
@@ -44,6 +44,7 @@ class ProtobufGenerator(
       )
       .indent
       .add(s"type EnumType = $name")
+      .add(s"override type EnumRecognizedType = ${e.recognizedEnum.fullName}")
       .print(e.getValues.asScala) {
         case (p, v) => p.add(s"def ${v.isName}: _root_.scala.Boolean = false")
       }
@@ -59,7 +60,7 @@ class ProtobufGenerator(
       }
       .add(s"object $name extends ${e.companionExtends.mkString(" with ")} {")
       .indent
-      .add(s"sealed trait ${e.recognizedEnum.nameSymbol} extends $name")
+      .add(s"sealed abstract class ${e.recognizedEnum.nameSymbol}(override val value: _root_.scala.Int) extends $name(value)")
       .add(s"implicit def enumCompanion: _root_.scalapb.GeneratedEnumCompanion[$name] = this")
       .print(e.getValues.asScala) {
         case (p, v) =>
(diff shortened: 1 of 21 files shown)


compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -59,7 +60,10 @@ class ProtobufGenerator(
       }
       .add(s"object $name extends ${e.companionExtends.mkString(" with ")} {")
       .indent
-      .add(s"sealed trait ${e.recognizedEnum.nameSymbol} extends $name")
+      .add(
+        s"sealed abstract class ${e.recognizedEnum.nameSymbol}(override val value: _root_.scala.Int) extends $name(value)"
```
thesamet, line 64: Is the change here necessary to implement `asRecognized`? (prefer not to break compatibility if not absolutely have to)

sideeffffect: you're right, I've amended the PR
maybe we can come back to this before the next breaking change release of ScalaPB

sideeffffect:
Hi @thesamet ! Have you had the change to look at this PR yet? :slightly_smiling_face:

thesamet:
Sorry for the delay. I am not comfortable with this proposal since inferred type for users' enum vals are going to become `Recognized` rather than `EnumName`.

sideeffffect:
I'm not sure I understand, this PR isn't changing anything about how the types are inferred, it just puts the `asRecognized` method to the `GeneratedEnum`, which is parent of all enum values.
But if you're concerned about the inferred type, maybe we could rename it from `Recognized` to `EnumNameRecognized`?

thesamet:
The concern is this:

```
val x = Seq(Weather.RAIN, Weather.SUNSHINE)
```

The inferred type of `x` would be `Seq[Weather.Recognized]`, or `Seq[WeatherRecognized]` (if we go with the proposal from the previous comment). This distinction would be unnecessary noise for most users.

sideeffffect:
I think I see what you mean. But I think that has been already fixed in this PR, no? The enum cases now don't inherit only from `Recognized` (as was the case when I opened the PR), I reverted this change, so things are back to normal, where enum cases inherit both from the parent sealed trait and Recognized, as before.
I updated the documentation to reflect that.

thesamet:
Closing due to inactivity.

sideeffffect:
Hello @thesamet , oh how time flies :sweat_smile: This must have slipped my mind.

Would you still be open to merging this if I revived the PR, incorporated the feedback, etc?

thesamet:
Hey @sideeffffect , time flies! If you're still interested, happy to review. Since it's been a while - it would be great if you can file a corresponding issue describing what problem we're here to solve, and what is the proposed solution. Also, is this change breaking in the sense that it would only go on the next release (0.12.x) ?

PR #812, opened 2020-04-22
Added StructUtils#fromStruct StructUtils#toStruct

ittaiz:
This PR adds the ability to convert from a `GeneratedMessage` to a `Struct` and back.
This is following a detailed discussion on gitter starting roughly [here](https://gitter.im/ScalaPB/community?at=5e864eb301c85847b2d0774f).

Few comments:
1. Feel free to nit pick, I don't work with scalatest so I might have made some non idiomatic things, I tried to follow `TextFormatSpec`.
2. I'm not sure what are your testing preferences but I went with fine-grain unit tests to allow to easily pinpoint bugs in the future.
3. I also added ser-deser tests which are duplicate of the "from"/"to". We can delete them if you'd like but I thought that seeing the e2e story has a lot of value as well.
4. I ended up adding `StructUtils` under scalapb-runtime/jvm and not shared because it needs classes from https://github.com/scalapb/ScalaPB/tree/master/scalapb-runtime/jvm/src/main/scala/com/google/protobuf/struct. If there is a way around it I'd love to know.
5. Most of the checks were on proto2 because that's the syntax of TestAllTypes. Any nuances important to check w.r.t. proto3? We use proto3 so it's especially important for us
6. One negative test I didn't add was in case someone sends a double value when the field is a float. In general the code is defensive because `Struct` has only one `Number` type as opposed to proto itself so I wanted to make sure people don't pass in float for int and so on. This specific case I ignored since the assumption is that most times people will use the serializer in any case. If you think it's important I'd be happy to add this test (though I'd love to know how you'd check that a double is not a float, compare against float max?).
7. I also added a few comments in the file itself for you to see and say if it's ok/we can do better.

Thanks!


ittaiz:
And I'm sorry the PR is hefty. The deserializer had a lot of edge cases...

e2e/src/main/protobuf/google/protobuf/fromstruct_two_level_nesting.proto
```diff
@@ -0,0 +1,56 @@
+// Protocol Buffers - Google's data interchange format
```
thesamet, line 1: Looks like this is an original file - so the copyright header is unnecessary.

ittaiz: 👍 would you like me to wait until you're done with the review before fixing?

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+import com.google.protobuf.struct.{ListValue, Struct, Value}
+import com.google.protobuf.struct.Value.Kind
+import scalapb.descriptors.{FieldDescriptor, PBoolean, PByteString, PDouble, PEmpty, PEnum, PFloat, PInt, PLong, PMessage, PRepeated, PString, PValue, ScalaType}
+
+import scala.util.{Failure, Success, Try}
+
+object StructUtils {
+  case class StructDeserError(index: Int, error: String)
+  def toStruct(generatedMessage: GeneratedMessage): Struct =
+  //TODO(@thesamet)- Can we ignore the case where generatedMessage is null?
```
thesamet, line 15: Yes, nulls are not handled throughout ScalaPB, which is idiomatic in Scala.

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+    toStruct(generatedMessage.toPMessage.value)
+
+  def fromStruct[T <: GeneratedMessage](
+                                         companion: GeneratedMessageCompanion[T],
```
thesamet, line 19: companion should be passed implicitly.

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+                                         struct: Struct
+                                       ): Either[StructDeserError, T] = {
+    //TODO(@thesamet)- do we want to fail (return Either#Left) if a structField arrives that we don't know about? I assume we don't want to fail on that for forward compatibility but want to double check
+    val fieldDescriptorToPValue = structMapToFDMap(companion, struct.fields)
+    fieldDescriptorToPValue.right.map(PMessage).map(companion.messageReads.read)
+  }
+
+  private def structMapToFDMap(companion: GeneratedMessageCompanion[_], structFields: Map[String, Value]): Either[StructDeserError, Map[FieldDescriptor, PValue]] = {
```
thesamet, line 27: For consistency with other code this should return `PMessage`. Looks like in all cases it's followed by wrapping the map in PMessage anyway.

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+  case class StructDeserError(index: Int, error: String)
+  def toStruct(generatedMessage: GeneratedMessage): Struct =
+  //TODO(@thesamet)- Can we ignore the case where generatedMessage is null?
+    toStruct(generatedMessage.toPMessage.value)
+
+  def fromStruct[T <: GeneratedMessage](
+                                         companion: GeneratedMessageCompanion[T],
+                                         struct: Struct
+                                       ): Either[StructDeserError, T] = {
+    //TODO(@thesamet)- do we want to fail (return Either#Left) if a structField arrives that we don't know about? I assume we don't want to fail on that for forward compatibility but want to double check
```
thesamet, line 22: I think that unknown fields should be silently ignored, similar to JSON. It could become a setting on v2 if needed.

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  private def toValue(pValue: PValue): Value = Value(pValue match {
+    case PInt(value) => Value.Kind.NumberValue(value.toDouble)
+    case PLong(value) => Value.Kind.StringValue(value.toString)
+    case PDouble(value) => Value.Kind.NumberValue(value)
+    case PFloat(value) => Value.Kind.NumberValue(value.toDouble)
+    case PString(value) => Value.Kind.StringValue(value)
+    case PByteString(value) => Value.Kind.StringValue(value.toStringUtf8) //TODO(@thesamet)- not sure if we have a better option here. WDYT?
+    case PBoolean(value) => Value.Kind.BoolValue(value)
+    case PEnum(value) => Value.Kind.StringValue(value.name.toString) //TODO(@thesamet)- value.name returns Any. Is toString ok?
```
thesamet, line 84: EnumValueDescription#name is a string - can you check why you're seeing Any?

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+    Struct(pMessageWithNonEmptyFields.map(pair => pair._1.name -> toValue(pair._2)))
+  }
+
+  private def toValue(pValue: PValue): Value = Value(pValue match {
+    case PInt(value) => Value.Kind.NumberValue(value.toDouble)
+    case PLong(value) => Value.Kind.StringValue(value.toString)
+    case PDouble(value) => Value.Kind.NumberValue(value)
+    case PFloat(value) => Value.Kind.NumberValue(value.toDouble)
+    case PString(value) => Value.Kind.StringValue(value)
+    case PByteString(value) => Value.Kind.StringValue(value.toStringUtf8) //TODO(@thesamet)- not sure if we have a better option here. WDYT?
```
thesamet, line 82: An arbitrary ByteStrings is most likely not a UTF-8 string, needs to be converted to Base64, check how it's done in scalapb-json4s.

e2e/src/test/scala/scalapb/StructUtilsSpec.scala
```diff
@@ -0,0 +1,344 @@
+package scalapb
```
thesamet, line 1: I appreciate the comprehensive testing here!

thesamet:
Done with initial pass. Will take another look once this set of changes are made. Thank you for putting this together!

ittaiz:
Done with all comments but the bytestring. Hope to nail that soon as well.

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
ittaiz: I think it's sbt/intellij issues. It shows red in IJ but builds fine 🤷

ittaiz:
@thesamet I'm done with the changes. Thanks for the prompt review :)

thesamet:
Overall looks good. Next step if for me to pull and try things locally, and also get to the bottom of this being jvm-only which would take a few days. It would be great if you can look into this in the meantime, the scala source files you indicated should be available also for js, just without java conversions.

ittaiz:
I think something is off with my IJ setup (all of the Any I saw) so I decided to just move the file via git move and see.
Indeed the build locally passed.
It's now under shared :)

ittaiz:
LMK if you need any more assistance from my side

ittaiz:
And also I'd love to know that proto3 doesn't need any more extra handling

ittaiz:
travis build is stuck.

ittaiz:
Finally green :)

scalapb-runtime/jvm/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,122 @@
...
+      struct: Struct
+  )(implicit companion: GeneratedMessageCompanion[T]): Either[StructDeserError, T] = {
+    val fieldDescriptorToPValue = structMapToFDMap(struct.fields)
+    fieldDescriptorToPValue.map(companion.messageReads.read)
+  }
+
+  private def structMapToFDMap(
+      structFields: Map[String, Value]
+  )(implicit companion: GeneratedMessageCompanion[_]): Either[StructDeserError, PMessage] = {
+    val fieldDescriptorToPValue = companion.scalaDescriptor.fields.map { fd =>
```
ittaiz, line 27: btw, I left a few Vals in place (here and above) because I like to make debugging easier. I can inline the vals if you prefer the full fluent idiom.

ittaiz:
@thesamet any thoughts? The merge isn’t very pressing to me since for now I copied over the code and it will take us a bit to upgrade but if there are feedbacks I’d rather fix them now when the context is fresh

ittaiz:
@thesamet gentle ping

thesamet:
@ittaiz sorry for the delay, You mentioned that you were able to move `StructUtils` to `shared`, but it's still in `jvm`, can you take a look?

ittaiz:
Somehow I didn’t push the last one. Please see now.

scalapb-runtime/shared/src/main/scala/scalapb/StructUtils.scala
```diff
@@ -0,0 +1,122 @@
...
+
+import com.google.protobuf.ByteString
+import com.google.protobuf.struct.Value.Kind
+import com.google.protobuf.struct.{ListValue, Struct, Value}
+import scalapb.descriptors._
+
+import scala.util.{Failure, Success, Try}
+
+object StructUtils {
+  case class StructDeserError(index: Int, error: String)
```
thesamet, line 13: Let's rename to `StructParsingError`. I think that having the index is not particularly useful since that single number doesn't help understand where the problem happened. I suggest that we use the `FieldDescriptor` instead.

ittaiz:
No problem about the rename.
Re field descriptor- not sure it will a lot more valuable than the number
since I try to give info in the message and it can add noise.
If you still want it I’d appreciate a bit of help in understanding how I
build one for the tests or alternatively use partial matching in scalatest
to ignore this member in the match

e2e/src/test/scala/scalapb/StructUtilsSpec.scala
```diff
@@ -0,0 +1,353 @@
...
+    val someFaultyInt = Random.nextFloat().toDouble
+    StructUtils.fromStruct(Struct(Map("optional_int32" -> Value(Value.Kind.NumberValue(someFaultyInt))))).left.value must be(
+      StructDeserError(1,s"Field `protobuf_unittest.TestAllTypes.optional_int32` is of type 'Int' but received 'NumberValue($someFaultyInt)'")
+    )
+  }
+
+  "fromStruct with faulty long field" should "pass" in {
+    val someFaultyLong = "Hi"
+    StructUtils.fromStruct(Struct(Map("optional_int64" -> Value(Value.Kind.StringValue(someFaultyLong))))).left.value must be(
+      StructDeserError(2,s"Invalid value for long: '$someFaultyLong'")
```
thesamet, line 312: This message doesn't show field names.

thesamet:
Let's consistently add the name of the field to all error messages, and then we don't really need the field number or field descriptor.

ittaiz:
Done

ittaiz:
I also aligned all error messages so they’re very similar

thesamet:
Thank you, @ittaiz !

ittaiz:
Thank you!!!
Any idea when you’ll be able to release?

ittaiz:
Hopefully with a new protoc dependency

PR #831, opened 2020-05-18
Suggest generating to a folder under sourceManaged

raboof:
To avoid the issue discussed at
https://github.com/thesamet/sbt-protoc/issues/92

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ containing the following line:
 Add the following line to your `build.sbt`:
 
     PB.targets in Compile := Seq(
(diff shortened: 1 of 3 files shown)


thesamet:
Yes, totally agreed. When this project started, this suggestion appeared to work better with IntelliJ afaik this is no longer the case.

raboof:
> When this project started, this suggestion appeared to work better with IntelliJ afaik this is no longer the case

I just did a quick test with IntelliJ and it indeed seems to work nowadays

PR #845, opened 2020-06-13
support custom service options in protoc-gen-scala

dsilva:
protoc-gen-scala currently fails on services with custom options like this:

```
--scala_out: java.lang.RuntimeException: Generated message class "com.google.protobuf.DescriptorProtos$ServiceOptions" missing method "getDeprecated".
```

Looks like that's because `make_reflect_config.sh` didn't exercise that class in the e2e protos.

Adding a custom service option in e2e/src/main/protobuf/service.proto for this.

### DIFF
--- a/e2e/src/main/protobuf/service.proto
+++ b/e2e/src/main/protobuf/service.proto
@@ -51,7 +51,13 @@ extend google.protobuf.MethodOptions {
     string custom_option = 50001;
 }
 
+extend google.protobuf.ServiceOptions {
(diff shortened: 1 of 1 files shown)


thesamet:
Merged! Thank you for reporting and investigating. Were you able to confirm that this addresses the issue?

dsilva:
Thank you!

protoc-gen-scala built from this branch worked with my proto files (with custom service options) that previously ran into this issue.

PR #939, opened 2020-10-09
add converters between WKT and their java.time counterparts

bjaglin:


scalapb-runtime/src/main/scala/scalapb/DurationCompanionMethods.scala
```diff
@@ -1,9 +1,19 @@
...
 
+import java.time.{Duration => jDuration}
+import scala.language.implicitConversions
+
 import com.google.protobuf.duration.Duration
 import com.google.protobuf.timestamp.Timestamp.NANOS_PER_SECOND
 
 trait DurationCompanionMethods {
+
+  implicit def fromJavaDuration(duration: jDuration): Duration =
```
thesamet, line 11: The common pattern I see in other libraries is that implicit conversions need to be explicitly imported by users, so that would be my preference.  Can this pair of methods and the other pair be added to some `DurationConverters` instead? Same for the other pair.

bjaglin: Indeed, implicit conversions should be explicit. Now I wonder if we couldn't have an `apply` on the companion object and a `toJavaTime` on the instances instead? Over the `DurationConverters` solution, that would have the benefit of making it more visible (IDE-wise). WDYT?

thesamet: I think either solution is fine, and even both (they aren't mutually exclusive)

bjaglin: Indeed. I went for both.

PR #946, opened 2020-10-13
Add an option to rename `oneof` fields

fmonniot:
This pull request add the option to rename a `oneof` field in a similar fashion as [renaming fields](https://scalapb.github.io/docs/customizations/#custom-names).

Here is the use case I'm trying to solve:

```proto
message Event {
    message Create {}
    message Other {
        // Will fail because notify is a reserved keyword
        oneof notify {
            // With this PR we can add `option (scalapb.oneof).scala_name = "myField";` to solve the issue
            Error error = 300;
        }
    }

    string id = 1;

    oneof change {
        Create create = 2;
        Other other = 8;
    }
}
```

As recommended [on Gitter](https://gitter.im/ScalaPB/community?at=5f84d1cda605930a78a618f9), this PR
- adds a `optional string scala_name` to `OneofOptions` in `scalapb.proto`,
- adapts `ExtendedOneofDescriptor#scalaName` to use `oneofOptions.getScalaName` if defined, and
- adds a test case `oneof_named_field.proto`, with a check in `OneofSpec.scala` (Not entirely certain if this is where the test should live though)


thesamet:
Looks great, no comments on the code. Will you be able to update
CHANGELOG.md and the docs at src/main/markdown/customizations.md ?

fmonniot:
Updated changelog and documentation

thesamet:
Looks great - thank you!

PR #977, opened 2020-11-15
Allow sealed oneofs to be defined inside a parent message

### LINKED ISSUE #974: Allow sealed oneofs to be part of a parent message
This is a carry-on from #751. Might be niche use-case, but it would be nice to be able to define a `preamble` for the body of a generated companion object. The idea is to have messages, as well as sealed traits, namespaced to an enclosing object.

My use case is Akka Typed, particularly with persistence, where there is a naming-convention to have all the messages (the protocol) for an actor inside a single object, including a `sealed trait Command` and `Event`, which is then extended by messages.

My current solution is to hack the `sealed trait` preamble into the companion object using `annotations`

```proto
message SomeProtocol {
  message State {
    // Hack to place the sealed traits in the SomeProtocol object
    option (scalapb.message).annotations = "sealed trait Command";
    option (scalapb.message).annotations = "sealed trait Event";

    // ...
  }

  message DoSomething {
    option (scalapb.message).extends = "Command";
    // ...
  }

  message SomethingDone {
    option (scalapb.message).extends = "Event";
    // ...
  }
}
``` 

This way `SomeProtocol.Command` and `SomeProtocol.DoSomething` signal that they belong to the same "protocol" and belong to each other by living in the same object. They are also easily and idiomatically imported together using `import SomeProtocol._`.

otto-dev:
Closes #974

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
@@ -50,9 +50,6 @@ class ProtoValidation(implicits: DescriptorImplicits) {
     m.getFields.asScala.foreach(validateField)
     if (m.isSealedOneofType) {
       val oneof = m.getRealOneofs.get(0)
-      if (m.getContainingType != null) {
-        throw new GeneratorException(s"${m.getFullName}: sealed oneofs must be top-level messages")
-      }
       if (m.getFields.size() != oneof.getFields.size()) {
         throw new GeneratorException(
           s"${m.getFullName}: sealed oneofs must have all their fields inside a single oneof"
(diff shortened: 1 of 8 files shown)


thesamet:
Nice, thanks for working on this! Can you also add a test to e2e named something like `sealed_oneof_nested.proto` that test the new scenario?

otto-dev:
Sure, let me have a look,

otto-dev:
Okay, I've added the tests. Some may be redundant, and maybe you have a better idea of how to structure the test-file. Please let met know! Thanks

thesamet:
LGTM - thanks for making this improvement!

otto-dev:
No, thanks to you @thesamet for this library and your excellent maintenance of the same!

Issue #998, opened 2020-12-08
scala.Array clashes with proto message Array in the same package


feoktant:
Hi,

I tried to implement [MySQL proto2 X Protocol](https://github.com/mysql/mysql-server/tree/8.0/plugin/x/protocol/protobuf). Source code was generating, but cannot be compiled. [Here is the repo to reproduce this issue](https://github.com/feoktant/scalapb-bug).

Main problem is that MySQL creates message `Array`([1](https://github.com/mysql/mysql-server/blob/8.0/plugin/x/protocol/protobuf/mysqlx_expr.proto#L319-L322),[2](https://github.com/mysql/mysql-server/blob/8.0/plugin/x/protocol/protobuf/mysqlx_datatypes.proto#L89-L91)).

Since in the same package case class `Array` was generated, it clashes like this:
```scala
  private lazy val ProtoBytes: Array[Byte] =
      scalapb.Encoding.fromBase64(scala.collection.immutable.Seq(....)
/*
com.mysql.cj.x.protobuf.mysqlx_expr.Array does not take type parameters
  private lazy val ProtoBytes: Array[Byte] =
*/

//-----------------
    com.google.protobuf.Descriptors.FileDescriptor.buildFrom(javaProto, Array(
      com.mysql.cj.x.protobuf.mysqlx_datatypes.MysqlxDatatypesProto.javaDescriptor
    ))
/*
type mismatch;
 found   : com.google.protobuf.Descriptors.FileDescriptor
 required: Seq[com.mysql.cj.x.protobuf.mysqlx_expr.Expr]
      com.mysql.cj.x.protobuf.mysqlx_datatypes.MysqlxDatatypesProto.javaDescriptor
*/
```

It could be easy fixed if instead of `Array` it will generate `scala.Array`

PR #1002, opened 2020-12-11
998: Fixed possible type clash

feoktant:
In case message is named `Array`, generated code will not compile.
Such situation happenend in MySQL X Protocol.

https://github.com/scalapb/ScalaPB/issues/998


thesamet:
Please run make_plugin_protos.sh to get the rest of the generated code updated. Thanks!

PR #1025, opened 2021-01-02
Scala 2.13.4

sullis:


thesamet:
Thanks for sending this. There are a few source incompatibilities in addition to the version bump that need to be fixed. I actually have it a local branch that I'll submit soon.

PR #1035, opened 2021-01-16
fix typos & group type-related features

bjaglin:


docs/src/main/markdown/validation.md
```diff
@@ -108,6 +108,30 @@ Validator[Person].validate(personInstance) match {
 }
 ```
 
+## Package-scoped extension options
```
bjaglin, line 111: this is not available outside preview, but it didn't really fit with the rest and the preview is going to be merged shortly so I thought it would be better up

thesamet:
Thank you!

PR #1054, opened 2021-01-31
scala-native

xuwei-k:


thesamet:
thank you!

thesamet:
Hey, looks like this is causing problems in the release part of the build. Can you help resolving this: https://github.com/scalapb/ScalaPB/runs/1819374106 ?

xuwei-k:
Hmmm, scala-library 2.12.13 and scala-collection-compat have same `@nowarn` annotation 😢 

- https://github.com/scala/scala/blob/v2.12.13/src/library/scala/annotation/nowarn.scala
- https://github.com/scala/scala-collection-compat/blob/v2.4.1/compat/src/main/scala-2.11_2.12/scala/annotation/nowarn.scala

Maybe we should use `MergeStrategy.first` or `MergeStrategy.last` 🤔 

https://github.com/sbt/sbt-assembly/tree/ec871177cc465188e1154852b160b969a532d006#merge-strategy

thesamet:
Either one, at runtime it probably won't matter which one we have.

xuwei-k:
https://github.com/scalapb/ScalaPB/pull/1057

PR #1055, opened 2021-02-01
Extend FieldMaskUtils functionality

evdokim:
FieldMaskUtils is currently missing several features from Java version such as:
* Ability to validate that field mask is formed correctly for a protobuf message
* Merge two or more field masks 
* Apply a field mask to a message

This PR aims to reimplement some of the missing methods in FieldMaskUtils. Original Java test cases were used when it was possible.

The following methods were added:

**applyToMessage**

Applies a field mask to a message.

```scala
def applyToMessage[M <: GeneratedMessage: GeneratedMessageCompanion](message: M): M
```

Java version:

```java
public static void merge(FieldMask mask, Message source, Message.Builder destination)
```

**fromFieldNumbers**

Constructs a field mask based on message numbers. Returns `None` when field number does not exist.

```scala
def fromFieldNumbers[M <: GeneratedMessage: GeneratedMessageCompanion](fieldNumbers: Int*): Option[FieldMask]
```

Java version:

```java
public static FieldMask fromFieldNumbers(Class<? extends Message> type, int... fieldNumbers)
```

**selectFieldNumbers**

A safer version of `fromFieldNumbers`, restrict to select only valid field numbers based on a predicate. Useful to define masks for all fields.

```scala
def selectFieldNumbers[M <: GeneratedMessage: GeneratedMessageCompanion](fieldNumberPredicate: Int => Boolean): FieldMask
```

**isValid**

Checks that field masks corresponds to a message schema,

```scala
def isValid[M <: GeneratedMessage: GeneratedMessageCompanion](fieldMask: FieldMask): Boolean
```

Java version:

```java
public static boolean isValid(Class<? extends Message> type, FieldMask fieldMask)
```

**containsFieldNumber**

Checks that field mask selects a certain field.

```scala
def containsFieldNumber[M <: GeneratedMessage: GeneratedMessageCompanion](fieldMask: FieldMask, fieldNumber: Int): Boolean
```

**union**

Unions two or more field masks.

```scala
def union(fieldMask: FieldMask, otherMasks: FieldMask*): FieldMask
```

Java version:

```java
public static FieldMask union(FieldMask firstMask, FieldMask secondMask, FieldMask... otherMasks)
```

### DIFF
--- a/scalapb-runtime/src/main/scala/scalapb/FieldMaskTree.scala
+++ b/scalapb-runtime/src/main/scala/scalapb/FieldMaskTree.scala
@@ -67,7 +67,7 @@ private[scalapb] case class FieldMaskTree(nodes: SortedMap[String, FieldMaskTree
   }
 }
 
-object FieldMaskTree {
+private[scalapb] object FieldMaskTree {
 
   val Empty: FieldMaskTree = FieldMaskTree(SortedMap.empty[String, FieldMaskTree])
 
--- a/scalapb-runtime/src/main/scala/scalapb/FieldMaskTree.scala
+++ b/scalapb-runtime/src/main/scala/scalapb/FieldMaskTree.scala
@@ -2,32 +2,44 @@ package scalapb
 
 import com.google.protobuf.Descriptors.{Descriptor, FieldDescriptor}
 import com.google.protobuf.field_mask.FieldMask
+import scalapb.FieldMaskTree.Empty
+import scalapb.descriptors.PMessage
 
 import scala.collection.SortedMap
 
 private[scalapb] case class FieldMaskTree(nodes: SortedMap[String, FieldMaskTree]) {
 
+  def applyToMessage[M <: GeneratedMessage: GeneratedMessageCompanion](message: M): M = {
+    val companion        = implicitly[GeneratedMessageCompanion[M]]
+    val filteredPMessage = applyToMessage(message.toPMessage)
+    companion.messageReads.read(filteredPMessage)
+  }
+
   def containsField[M <: GeneratedMessage: GeneratedMessageCompanion](fieldNumber: Int): Boolean = {
-    val descriptor = implicitly[GeneratedMessageCompanion[M]].javaDescriptor
-    Option(descriptor.findFieldByNumber(fieldNumber)) match {
-      case Some(field) => nodes.contains(field.getName)
+    val descriptor = implicitly[GeneratedMessageCompanion[M]].scalaDescriptor
+    descriptor.findFieldByNumber(fieldNumber) match {
+      case Some(field) => nodes.contains(field.name)
       case None        => false
     }
   }
 
   def fieldMask: FieldMask = FieldMask(paths)
 
-  private def paths: Vector[String] = {
-    nodes.toVector.flatMap {
-      case (field, FieldMaskTree.Empty) => Vector(field)
-      case (field, subTree) =>
-        subTree.paths.map(path => s"$field${FieldMaskTree.FieldSeparator}$path")
-    }
-  }
-
   def isValidFor[M <: GeneratedMessage: GeneratedMessageCompanion]: Boolean = {
     val descriptor = implicitly[GeneratedMessageCompanion[M]].javaDescriptor
     isValidFor(descriptor)
+  }
+
+  private def applyToMessage(message: PMessage): PMessage = {
+    val fieldValues = message.value.keys.flatMap { fieldDescriptor =>
+      val value = (message.value.apply(fieldDescriptor), nodes.get(fieldDescriptor.name)) match {
+        case (pValue, Some(Empty))                 => Some(pValue)
+        case (subMessage: PMessage, Some(subTree)) => Some(subTree.applyToMessage(subMessage))
+        case _                                     => None
+      }
+      value.map(fieldDescriptor -> _)
+    }.toMap
+    PMessage(fieldValues)
   }
 
   private def isValidFor(descriptor: Descriptor): Boolean = {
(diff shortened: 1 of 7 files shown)


scalapb-runtime/src/main/scala/scalapb/FieldMaskTree.scala
```diff
@@ -0,0 +1,108 @@
...
+  private def paths: Vector[String] = {
+    nodes.toVector.flatMap {
+      case (field, FieldMaskTree.Empty) => Vector(field)
+      case (field, subTree) =>
+        subTree.paths.map(path => s"$field${FieldMaskTree.FieldSeparator}$path")
+    }
+  }
+}
+
+object FieldMaskTree {
```
thesamet, line 70: should this be marked `private[scalapb]` as well, since the class is package private too?

evdokim: Yes, makes sense

scalapb-runtime/src/test/scala/scalapb/FieldMaskTreeSpec.scala
```diff
@@ -0,0 +1,34 @@
+package scalapb
+
+import com.google.protobuf.field_mask.FieldMask
+import munit.FunSuite
+
+class FieldMaskTreeSpec extends FunSuite {
+  test("testAddFieldPath") {
+    // https://github.com/protocolbuffers/protobuf/blob/v3.6.0/java/util/src/test/java/com/google/protobuf/util/FieldMaskTreeTest.java#L39-L62
```
thesamet, line 8: The version used for testing is fairly outdated, can you catch up the tests with the 3.14.x branch?

evdokim: Yes, I noticed it, but didn't want to touch other tests in `FieldMaskUtilsSpec`. I will update all them to `3.14` then

thesamet:
This looks great, thanks for contributing it! Can you add javaodcs for all public methods. One-liners such as the ones you used in the [comment above](https://github.com/scalapb/ScalaPB/pull/1055#issue-565308072) would work.

evdokim:
> This looks great, thanks for contributing it! Can you add javaodcs for all public methods. One-liners such as the ones you used in the comment above would work.

Thank you, I am happy that I could contribute. I added some docs.

thesamet:
Are there additional test cases from 3.14.0 that need to be added?

evdokim:
> Are there additional test cases from 3.14.0 that need to be added?

I didn't noticed any new cases there

thesamet:
In 3.14.0, protobuf-java have added [testMergeRequire](https://github.com/protocolbuffers/protobuf/blob/v3.14.0/java/util/src/test/java/com/google/protobuf/util/FieldMaskTreeTest.java#L174) which I believe should be added. There are few more tests for functionality that doesn't exist yet in Scala (subtract, removeFieldPath) which is ok to ignore for this PR.

evdokim:
> In 3.14.0, protobuf-java have added testMergeRequire which I believe should be added.

I just tried to port this test, but I am not sure whether it makes so much sense in context of ScalaPB. The thing is that it operates with required fields and as I understand they behave quite differently with Java builders. For example it is possible to have an instance of builder where certain required fields are not specified ([see](https://github.com/protocolbuffers/protobuf/blob/v3.14.0/java/util/src/test/java/com/google/protobuf/util/FieldMaskTreeTest.java#L188)), whereas in Scalapb we work with case classes that where all fields are defined.

> There are few more tests for functionality that doesn't exist yet in Scala (subtract, removeFieldPath) which is ok to ignore for this PR.

True, in this PR I ported only the functionality that we are currently use, but original Java class has a couple of more things. I am happy to extend it later if there is a need.

thesamet:
In the Scala case, is it possible when applying a mask to omit a required field, and then expect `applyToMessage` to throw?

evdokim:
> In the Scala case, is it possible when applying a mask to omit a required field, and then expect applyToMessage to throw?

Yes, this is exactly the second case in `testMergeRequire`, I managed to port it without much changes

thesamet:
Fantastic - thank you, @evdokim

jroper:
This was missing merge support, which is necessary for implementing patch functionality - that is, take an object M1 that serves as a patch, and copy the fields in it specified by a FieldMask over to M2, leaving all other fields in M2 unchanged.

thesamet:
@jroper , I think this be accomplished by something like `M2.mergeFrom(fieldMaskTask.applyToMessage(M1))`. The behavior of this need to be looked at: for instance, repeated fields will be appended to, and we need to ensure nested messages don't get completely replaced but patched. Ideally what we do should follow the same semantics of protobuf-java implementation. If needed, can you test if this snippet works and otherwise open a new issue with a feature request?

jroper:
I can't see a `mergeFrom` method, ScalaPb doesn't generate one does it? I think `mergeFrom` is a method on the Java generated `Builder` classes.

thesamet:
Sorry, my memory failed me! We don't have `mergeFrom`. There is a hacky way though that does the same: `cmp.parseFrom(m1.toByteArray ++ m2.toByteArray)`...

jroper:
Hmm.... that's a good point. This explains why the update logic is as it is, with repeated fields being concatenated, because that's what the protobuf wire encoding requires.

jroper:
Actually, no, I'm wrong. The thing is, merging that way won't allow reseting a field to it's default, here's what the spec says:

> In order to reset a field's value to the default, the field must be in the mask and set to the default value in the provided resource.

So, neither `mergeFrom` nor message concatenation will achieve that, since fields set to their default value are ignored in `mergeFrom`, and ignored when you convert the message to a byte array. It does require specific support for `FieldMask` based merging to work according to the spec.

thesamet:
Yes, looks like additional code is necessary. Maybe a utility function that flips all the scalar (non repeating, not messages) fields in the mask can be used to clear `M2` before patching using the byte concatenation. But yes - this needs more thought and tests before confirming it's up to spec.

PR #1073, opened 2021-02-22
Allow negative nanos of Duration

### LINKED ISSUE #1068: Duration can not have negative nanos
In ScalaPB 0.10.11, negative nanos field of Duration is not valid.

```scala
import com.google.protobuf.duration.Duration
import scala.math.Ordered.orderingToOrdered
Duration(0, 0) > Duration(-1, -2)
```

```
requirement failed: Duration Duration(-1,-2,UnknownFieldSet(Map())) is not valid.
java.lang.IllegalArgumentException: requirement failed: Duration Duration(-1,-2,UnknownFieldSet(Map())) is not valid.
```

According to Google's specification, nanos can actually be a negative value.

> Signed fractions of a second at nanosecond resolution of the span of time. Durations less than one second are represented with a 0 `seconds` field and a positive or negative `nanos` field. For durations of one second or more, a non-zero value for the `nanos` field must be of the same sign as the `seconds` field. Must be from -999,999,999 to +999,999,999 inclusive.

https://github.com/protocolbuffers/protobuf/blob/v3.14.0/src/google/protobuf/duration.proto

If it is a bug, I will create a pull request to fix [DurationCompanionMethods.scala](https://github.com/scalapb/ScalaPB/blob/master/scalapb-runtime/src/main/scala/scalapb/DurationCompanionMethods.scala).

piyo7:
Fix #1068

> Note that if we allow negative nanos, the Duration would have to be normalized before being compared for ordering. There are reference implementation here: https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/TimeUtil.java and https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/Durations.java

@thesamet 
Thanks for your comment.
I think that validator does not need to call normalizer just as `Durations.isValid` does not call `TimeUtil.normalizedDuration` in Durations.java. Normalizer is required by math operators like `Durations.add` and `Durations.subtract`.
Although, I want math operators for Duration and Timestamp. I'm wondering if creating another pull request to implement them as protobuf-java-util wrapper.

### DIFF
--- a/e2e/src/test/scalajvm/TimestampsSpec.scala
+++ b/e2e/src/test/scalajvm/TimestampsSpec.scala
@@ -38,4 +38,20 @@ class TimestampsSpec extends AnyFlatSpec with Matchers {
         duration must be(Duration(1, 333000000))
         duration.asJavaDuration must be(jtDuration.between(t1, t2))
     }
+
+    "Durations" should "be valid when their seconds and nanos are in defined range and have same sign" in {
+        Duration.isValid(Duration(315576000001L, 0)) must be(false)
+        Duration.isValid(Duration(315576000000L, 999999999)) must be(true)
+        Duration.isValid(Duration(315576000000L, 0)) must be(true)
+        Duration.isValid(Duration(315576000000L, -999999999)) must be(false)
+        Duration.isValid(Duration(0, 1000000000)) must be(false)
+        Duration.isValid(Duration(0, 999999999)) must be(true)
+        Duration.isValid(Duration(0, 0)) must be(true)
+        Duration.isValid(Duration(0, -999999999)) must be(true)
+        Duration.isValid(Duration(0, -1000000000)) must be(false)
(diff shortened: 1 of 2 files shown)


thesamet:
Without normalizing, wouldn't (1,-999999999) be larger than (0,2) ?

piyo7:
`Duration(1,-999999999)` is not valid. This should be represented with a 0 seconds field.

> Durations less than one second are represented with a 0 seconds field and a positive or negative nanos field. 

https://github.com/protocolbuffers/protobuf/blob/v3.14.0/src/google/protobuf/duration.proto

Therefore, I think `Duration(1,-999999999) < Duration(0,2)` should throw IllegalArgumentException from checkValid.

thesamet:
Thanks for clarifying, I missed the requirement that the signs of the nanos and seconds must not mismatch.

PR #1127, opened 2021-04-19
Tolerate a message called `Unit` when generating gRPC services. 

### LINKED ISSUE #1126: Generated gRPC code compilation fails on messages called Unit
Similar to #427, but for a different area of code. 

The code generated by `GrpcServicePrinter` references `Unit` ambiguously. When the service in question uses a message called `my.custom.Unit`, then the `invoke` methods on the generated `ServerCalls` instances end up referring to the `my.custom.Unit` instead of `scala.Unit`.

I expect this can be fixed by replacing occurences of `: Unit` with `: _root_.scala.Unit` in `GrpcServicePrinter`.

Grogs:
Fixes #1126. The addition of `message Unit {}` to `src/main/protobuf/service.proto` surfaces the issue. You can revert the changes to  `GrpcServicePrinter` to see the error.


thesamet:
Thank you.

PR #1156, opened 2021-05-24
feature/enum-annotations: Annotations for enum

matepek:
(In progress...)

I was thinking using the `private` as an annotation but I don't like the idea. It's not for that and values will be annotated using 2 lists so the second one should have the `private` or the first but then the second has to be empty. Of course I could switch them but thats not a solution and the problem will be the same.


docs/src/main/markdown/customizations.md
```diff
@@ -721,3 +721,10 @@ message BarMessage {
 }
 ```
 
+In ScalaPB 0.11.2, you can also add annotations to the enum values and the `Unrecognized` case class:
```
thesamet, line 724: 0.11.4

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,18 @@ message EnumOptions {
   // must be present.
   optional string type = 3;
 
+  // Custom annotations to add to the generated case object.
+  repeated string annotations_for_trait = 4;
+
+  // Custom annotations to add to the generated case object.
+  repeated string annotations_for_value = 5;
+
+  // Custom annotations to add to the generated Unrecognised case class.
```
thesamet, line 300: typo: Unrecognized :)

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,18 @@ message EnumOptions {
   // must be present.
   optional string type = 3;
 
+  // Custom annotations to add to the generated case object.
+  repeated string annotations_for_trait = 4;
```
thesamet, line 295: For consistency with the existing `companion_annotations` name, please rename to`trait_annotations` - same comment apply to the new fields below.

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,18 @@ message EnumOptions {
   // must be present.
   optional string type = 3;
 
+  // Custom annotations to add to the generated case object.
```
thesamet, line 294: s/generated case object/generated enum's base trait/

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,18 @@ message EnumOptions {
...
+  repeated string annotations_for_trait = 4;
+
+  // Custom annotations to add to the generated case object.
```
thesamet, line 297: I think it should be "all generated case objects"? Is it including or excluding the Unrecognized - let's specify.

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,18 @@ message EnumOptions {
...
+  repeated string annotations_for_value = 5;
+
+  // Custom annotations to add to the generated Unrecognised case class.
+  repeated string annotations_for_unrecognized = 6;
+
+  // If true, the generated Unrecognized case class will be private.
```
thesamet, line 303: I prefer to avoid having specific presets for consistency with the other annotations. There are different types of private in Scala (private and package-private) and different users might want different things.

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 ## [0.11.2] (Unreleased)
 - Fix bug where generated code would not compile if a message was named Unit (#1126)
 - Add option to block generation/access to field getters (#1098)
-- 
+- Add option for enum annotation (#1155)
```
thesamet, line 5: Our changelog is out of date. This should go into a new "0.11.4 (Unreleased)" section.

thesamet, line 5: options for enum annotations

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -815,6 +815,24 @@ class DescriptorImplicits private[compiler] (
...
+        Nil
+    }
+
+    def annotationForTraitList: Seq[String] =
+      deprecatedAnnotation ++ scalaOptions.getAnnotationsForTraitList().asScala.toSeq
+
+    def annotationForValueList: Seq[String] =
+      deprecatedAnnotation ++ scalaOptions.getAnnotationsForValueList().asScala.toSeq
+
+    def annotationListForUnrecognised: Seq[String] =
```
thesamet, line 832: unrecognizedAnnotations

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -815,6 +815,24 @@ class DescriptorImplicits private[compiler] (
...
+      deprecatedAnnotation ++ scalaOptions.getAnnotationsForUnrecognizedList().asScala.toSeq
+
+    def isPrivateUnrecognized: Boolean = scalaOptions.getUnrecognizedIsPrivate
```
thesamet, line 835: Let's remove (see rational in another comment)

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -815,6 +815,24 @@ class DescriptorImplicits private[compiler] (
...
     }
+
+    private[this] def deprecatedAnnotation: Seq[String] = {
+      if (enumDescriptor.getOptions.getDeprecated)
+        List(ProtobufGenerator.deprecatedAnnotation)
+      else
+        Nil
+    }
+
+    def annotationForTraitList: Seq[String] =
```
thesamet, line 826: traitAnnotations

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -815,6 +815,24 @@ class DescriptorImplicits private[compiler] (
...
+      deprecatedAnnotation ++ scalaOptions.getAnnotationsForTraitList().asScala.toSeq
+
+    def annotationForValueList: Seq[String] =
```
thesamet, line 829: valueAnnotations

protobuf/scalapb/scalapb.proto
```diff
@@ -309,6 +321,9 @@ message EnumValueOptions {
   // Name in Scala to use for this enum value.
   optional string scala_name = 2;
 
+  // Custom annotations to add to the generated case object.
```
thesamet, line 324: "to the generated case object for this enum vlaue".

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
thesamet, line 829: Shouldn't this method take the EnumValueDescriptor too so it concats the annotations from the EnumValueDescriptor's options?

thesamet:
Thanks for working on this PR. Comments are mostly around naming consistency.  Yes, let's add tests too.

thesamet (changes requested):

protobuf/scalapb/scalapb.proto
matepek: I'm moving this conversation to the main thread

matepek:
@thesamet:
> I prefer to avoid having specific presets for consistency with the other annotations. There are different types of private in Scala (private and package-private) and different users might want different things.

Consider the followings:
- This is a library. Its inner business that something is private or not. With this feature we are just encourage the user to use it correctly.
- Having this private option: the user still can disable private and add some "magic" annotation like `private[package]`
- As far as I understand the intended usage of this lib the users shouldn't create `Unrecognized` values by themselves. My issue shows exactly why. I don't see much point exposing it to them.
- Once again, (because i think this is the most important point) encouraging the user to to use the lib correctly. 

(Maybe I would also consider making it private by default in the next major version.)

What do you think?

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
matepek: I was debating pros and cons and ended up leaving like this because the reader can see it better whats happening in the generator:
```
          .add("@SerialVersionUID(0L)")
          .seq(e.valueAnnotationList)
          .seq(v.annotationList)
```

Another thing I was thinking: Is there a scenario when the order of the annotations really matter? Shouldn't right?

thesamet:
> What do you think?
Overall, I think that making it private could have been a valid design choice when `Unrecognized` were introduced. Making it private now, either in opt-in or opt-out mode, comes at a cost to users if its opt-out and to maintainers if it's opt-in.

For users: the current legit use case by users is pattern matching. Users handle the unrecognized case by matching on `_: Unrecognized`, marking it private by default in the future would break user code. Also, test code out there is instantiating using `Unrecognized` to test scenarios. I agree that it should be preferred to instantiate using `fromValue`.

For maintainers: every option added comes with maintenance cost. Adding an explicit `unrecognized_is_private` option communicates to users that this is a supported and tested combination and it generally gets maintained long term. Given that the feedback you provided on `Unrecognized` constructor being public hasn't been voiced so far, I prefer at this point to not take on supporting it. However, my goal is to provide value to users through this library, so I am willing to reconsider in the future.

matepek:
Pattern matching of course... Good point.

Hm... How about this?

```
   s"""final case class Unrecognized private[$name](unrecognizedValue: _root_.scala.Int) extends $name(unrecognizedValue) with _root_.scalapb.UnrecognizedEnum
           |object Unrecognized {
           |  def apply(unrecognizedValue: _root_.scala.Int): $name = fromValue(unrecognizedValue) 
           |}
           |lazy val values = scala.collection.immutable.Seq(${e.getValues.asScala
          .map(_.scalaName.asSymbol)
          .mkString(", ")})
           |def fromValue(__value: _root_.scala.Int): $name = __value match {""".stripMargin
      )
      .print(e.valuesWithNoDuplicates) { case (p, v) =>
        p.add(s"  case ${v.getNumber} => ${v.scalaName.asSymbol}")
      }
      .add(
        s"""  case __other => new Unrecognized(__other)
           |}
```

thesamet:
I think it's a fairly reasonable change, my earlier comment still applies - though I want to refrain from breaking public API or maintaining an opt-in feature that needs to be tested/supported in the long run.

matepek:
Your concerns are valid. I just believe that this situation should be handled. But the newly proposed one is actually better than the one I previously suggested.

I added `@deprecated` annotation to the Unrecognized.apply.
Because of this change the annotations are not necessary for the purpose but I can continue it.

If there are no further suggestions I will finish it soon.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -50,33 +50,41 @@ class ProtobufGenerator(
...
-        .map(_.scalaName.asSymbol)
-        .mkString(", ")})
-              |def fromValue(__value: _root_.scala.Int): $name = __value match {""".stripMargin)
+      .add("@SerialVersionUID(0L)")
+      .seq(e.unrecognizedAnnotationList)
+      .add(
+        s"""final case class Unrecognized private[$name](unrecognizedValue: _root_.scala.Int) extends $name(unrecognizedValue) with _root_.scalapb.UnrecognizedEnum
+           |object Unrecognized {
+           |  @deprecated("Could have lead to issues before. Use $name.fromValue instead. This might be private in the future.")
+           |  def apply(unrecognizedValue: _root_.scala.Int): $name = fromValue(unrecognizedValue)
```
thesamet, line 76: That's a great idea, but unfortunately it breaks binary compatibility (see the mima test). I suggest that the changes around this (deprecation and making the constructor private) are split to a different PR which we can target for next stable (0.12 or 1.0)

CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Change Log
-## [0.11.2] (Unreleased)
+
+## [0.11.4] (Unreleased)
+- Add options for enum annotations (#1155)
+- Add `unrecognized_value_class_is_private` option to make `Unrecognised` class private for safer development. (#1155)
```
thesamet, line 5: This line needs to be updated.

matepek:
Hey there,
So I couldn't figure out in a reasonable time how to write a test which checks the annotation of a scala object so that part is missing (see TODO), otherwise It might ready. pls check.

protobuf/scalapb/scalapb.proto
```diff
@@ -291,6 +291,15 @@ message EnumOptions {
   // must be present.
   optional string type = 3;
 
+  // Custom annotations to add to the generated enum's base class.
+  repeated string base_annotations = 4;
+
+  // Custom annotations to add to the generated trait.
+  repeated string recognised_annotations = 5;
```
thesamet, line 298: recognized

docs/src/main/markdown/customizations.md
```diff
@@ -721,3 +721,10 @@ message BarMessage {
 }
 ```
 
+In ScalaPB 0.11.4, you can also add annotations to the enum values and the `Unrecognized` case class:
+TODO:mapek
```
thesamet, line 725: Can you fix the example? Once merged this gets automatically deployed to the live site.

thesamet:
Reviewed - left some comments. Sorry for the delay, was away and missed the notifications.

protobuf/scalapb/scalapb.proto
matepek: sure

docs/src/main/markdown/customizations.md
matepek: sure

PR #1170, opened 2021-06-03
add NotPreserveUnknownFields as GeneratorOption

alex-gaiduchok:
add NotPreserveUnknownFields as GeneratorOption

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/GeneratorOption.scala
+++ b/compiler-plugin/src/main/scala/scalapb/GeneratorOption.scala
@@ -30,4 +30,8 @@ object GeneratorOption {
   case object RetainSourceCodeInfo extends GeneratorOption {
     override def toString = "retain_source_code_info"
   }
+
+  case object NotPreserveUnknownFields extends GeneratorOption {
+    override def toString = "not_preserve_unknown_fields"
+  }
 }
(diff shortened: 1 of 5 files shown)


thesamet:
Hi @alex-gaiduchok , thanks for sending this PR. We are refraining from adding new generator options that change the public API of the generated code since it breaks compatibility between ScalaPB plugins and leads to confusing developer experience. See this [FAQ](https://scalapb.github.io/docs/faq#why-a-certain-customization-is-not-available-as-a-global-generator-parameter) for more details. The recommended alternative for users is to use package-scoped options.

alex-gaiduchok:
@thesamet thank you, was waiting for the response in gitter meanwhile added this draft) tbh its gonna be painful in my case cause proto files used not only by scala generator

thesamet:
@alex-gaiduchok , in SBT you can add an additional directory to PB.protoSources just for the package-scoped options file, so your main protos directory isn't affected and other tools will not see it:

```
Compile / PB.protoSources += baseDirectory.value / "src"/ "main" / "scalapb-options"
```

alex-gaiduchok:
> @alex-gaiduchok , in SBT you can add an additional directory to PB.protoSources just for the package-scoped options file, so your main protos directory isn't affected and other tools will not see it:
> 
> ```
> Compile / PB.protoSources += baseDirectory.value / "src"/ "main" / "scalapb-options"
> ```
> 
> Then place the package-scoped options file in `src/main/scalapb-options`.

Thanks, I'll give it a try =)

PR #1230, opened 2021-09-14
Add new Scala 3 reserved keywords

### LINKED ISSUE #1226: Update resvered keyword list for Scala 3
There are a few new reserved keywords in Scala 3, see list here: https://dotty.epfl.ch/docs/internals/syntax.html#keywords. The generated scala code can be invalid if it contains any of these. One of our schemas is using the term `open` in an enum, resulting in a syntax error.

Updating https://github.com/scalapb/ScalaPB/blob/428cd7139ed3e48566ee43339524bd41fcae9732/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala#L1106 might  resolve this issue. 

Thanks!

rgueldem:
Add missing [reserved keywords](https://dotty.epfl.ch/docs/internals/syntax.html#keywords) for Scala 3.
Fixes #1226.

- [x] Add e2e coverage for new keywords


thesamet:
I think soft keywords and modifiers are not necessary, a definition like `val opaque: Int = 3` compiles on Scala 3.0.2. Have you checked if this leads to a compiler error? There is a bit of user experience reduction if used too sparingly for common words (open, transparent, inline)

rgueldem:
I was adjusting this PR to only add the new regular keyword (`given`) and the soft modifiers. The original issue I ran into was with `open`, I have not yet tried the other modifiers.

rgueldem:
The problematic usage was in a pattern match:
```
__value match {
    case 0 => UNKNOWN_CONVERSATION_STATUS
    case 1 => open
    case 2 => pending
    case 3 => solved
    case __other => Unrecognized(__other)
  }
```

rgueldem:
Minimal example that reproduces the issue https://scastie.scala-lang.org/WWFbz0tWRvC1TYOmB1OG1w

rgueldem:
One additional point, the problematic proto file on our end is not following the [style guide](https://developers.google.com/protocol-buffers/docs/style#enums) for enums and uses lower case for value names. Out of a few 100 proto files in our project, only a handful use that style, so it's probably not a widespread problem. Let me know if you are still interested in addressing it in scalapb. I can also see if we are able to roll out breaking changes to those proto files on our end to get them compliant with the style guide which would fix the compilation error as well...

erwan:
👋  any news on this PR, is it only waiting for review?

thesamet:
Thanks!

PR #1241, opened 2021-10-01
POC 2.11 runtime build

borissmidt:
Dear, 

We really like the custom type feature of scalapb but our build has become overly complex because the lack of scala 2.11 support and the incompatible code generation between 0.9 and 0.11. 

So i tried cross-building the scalapb runtime to 2.11 and it isn't that hard to add support.

Are you interested in this feature? 
Was there any specific reason to drop 2.11 support?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -159,7 +159,7 @@ lazy val compilerPlugin = (projectMatrix in file("compiler-plugin"))
     Compiler.generateVersionFile,
     Compiler.generateEncodingFile
   )
-  .jvmPlatform(scalaVersions = jvmPlatformVersions)
+  .jvmPlatform(scalaVersions = Seq(Scala212, Scala213))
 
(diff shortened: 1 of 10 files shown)


thesamet:
Thanks for sending the PR. Scala 2.11 reached End Of Life around 2018, and to reduce maintenance, development and testing efforts, support for Scala 2.11 has been dropped. We're already maintaining the last three major versions (2.12, 2.13, 3.0) across JVM, Scala.js and Scala Native whenever possible. The latest version of some of the libraries we are using (scalacheck for instance) also dropped Scala 2.11 support, so we will not be able to test our build if we bring back Scala 2.11.  

If your project is on Scala 2.11, is there any reason not to keep using ScalaPB 0.9.x? If there's any feature missing in ScalaPB 0.9.x, we can look into backporting it.

PR #1278, opened 2021-12-09
Update protobufCompilerVersion to 0.19.1

jendakol:
Hi, we've found out ScalaPB uses quite an old version of the Protobuf compiler. Unless there's a strong reason why it does so, please accept this PR with a version update (and release a new ScalaPB version after that 😇).  
Thanks!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import com.typesafe.tools.mima.core._
 import BuildHelper._
 import Dependencies._
 
-val protobufCompilerVersion = "3.15.8"
+val protobufCompilerVersion = "3.19.1"
 
 val MimaPreviousVersion = "0.11.0"
 


thesamet:
This impacts the version the compiler users internally, not the version user projects are getting at runtime (that's in https://github.com/scalapb/ScalaPB/blob/master/project/Dependencies.scala#L9). The version used by protoc is in a different project (sbt-protoc), and there's a PR for updating it but it requires some fixes.

jendakol:
Oh 😄 OK 🤔 We want to use 3.19.1 in our project and there was a breaking change in the dependency between 3.17 and 3.18, so using it together with the current version of the Gradle Google protobuf plugin makes it explode. Bad story. If we update the dependency of Protobuf java to 3.19..1 (from the 3.15.8 being transitive dep. from your project) it _seems_ to work. Do you think it's a way to hell and we should just accept what we get from you or it's safe to just override the dependency?

thesamet:
I released ScalaPB 0.11.7 and sbt-protoc 1.0.5, both with protobuf/protoc 3.19.1.

jendakol:
Thank you!

PR #1307, opened 2022-01-18
Revert Scala 3 from 3.1.0 to 3.0.2

jodersky:
As the comment says, 3.x is not forwards compatible. Since ScalaPB is a library used by many others, if possible, a conservative scala 3 version should be preferred.


jodersky:
See also: https://github.com/scala-steward-org/scala-steward/issues/2467

PR #1331, opened 2022-02-24
Fix missing import definitions in services.

### LINKED ISSUE #1332: Imports stated in transformers are not added to grpc classes.
when i include transformers like:
it seems that the models pick up the imports of the converters properly but any service generated source file does not pick up this import statement.

package com.restore
option (scalapb.options) = {
  scope: PACKAGE
  import: "com.restore.protobuf.ProtobufConversions.{durationToProto, periodToProto, frequencyToProto, powerToProto, energyToProto, intervalToJoda, timestampToJoda}"

so in the generated sources i see in all source under of package com.restorean

import com.restore.protobuf.ProtobufConversions.{durationToProto, periodToProto, frequencyToProto, powerToProto, energyToProto, intervalToJoda, timestampToJoda}

but for the Service files (ending with Grpc) the import is missing causing my build to fail.

i'm using the 11.9 version of scalapb

borissmidt:
The         scalaFileHeader was not called for grpc services. 
fixes #1332

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
@@ -251,7 +251,6 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
   def printService(printer: FunctionalPrinter): FunctionalPrinter = {
     printer
       .add(
-        "package " + service.getFile.scalaPackage.fullName,
         "",
         s"${service.deprecatedAnnotation}object ${service.companionObject.nameSymbol} {"
       )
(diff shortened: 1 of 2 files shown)


compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -251,7 +251,6 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
   def printService(printer: FunctionalPrinter): FunctionalPrinter = {
     printer
       .add(
-        "package " + service.getFile.scalaPackage.fullName,
```
borissmidt, line 254: this is already set by the Fileheader.

borissmidt:
Thank you for the fast response, are there any releases planned?

PR #1336, opened 2022-02-28
Add namespaces for other supported languages

### LINKED ISSUE #1334: Add package names for other languages
Hi there!

We've been using zio-grpc to build an API that we intend to expose to our clients, with the intent to publish the IDL so they can generate their own API clients. I tried building a go example but it failed because the scalapb proto file did not have a `go_package` option.

Could the following options be added: `csharp_namespace`, `go_package`, `php_namespace` and `ruby_package`?

Happy to raise a PR myself, just need to know what values would make sense?

regiskuckaertz:
Closes #1334 

I ran the `make_plugin_proto` script but no difference, as expected I guess 😄

### DIFF
--- a/protobuf/scalapb/scalapb.proto
+++ b/protobuf/scalapb/scalapb.proto
@@ -2,7 +2,7 @@ syntax = "proto2";
 
 package scalapb;
 
-option go_package = "scalapb.github.io/cda/grpc/trades";
+option go_package = "scalapb.github.io/protobuf/scalapb";
 option java_package = "scalapb.options";
 
 option (options) = {
--- a/protobuf/scalapb/scalapb.proto
+++ b/protobuf/scalapb/scalapb.proto
@@ -2,7 +2,6 @@ syntax = "proto2";
 
 package scalapb;
 
-option csharp_namespace = "ScalaPb.Options";
 option go_package = "scalapb.github.io/cda/grpc/trades";
 option java_package = "scalapb.options";
 
--- a/protobuf/scalapb/scalapb.proto
+++ b/protobuf/scalapb/scalapb.proto
@@ -3,9 +3,8 @@ syntax = "proto2";
 package scalapb;
 
 option csharp_namespace = "ScalaPb.Options";
-option go_package = "scalapb.github.io/pirum.com/cda/grpc/trades";
+option go_package = "scalapb.github.io/cda/grpc/trades";
 option java_package = "scalapb.options";
-option php_namespace = "ScalaPb\\Options";
 
 option (options) = {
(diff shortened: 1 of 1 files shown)


protobuf/scalapb/scalapb.proto
```diff
@@ -2,7 +2,10 @@ syntax = "proto2";
 
 package scalapb;
 
+option csharp_namespace = "ScalaPb.Options";
+option go_package = "scalapb.github.io/pirum.com/cda/grpc/trades";
```
thesamet, line 6: The package name here seem to contain identifiers specific to your company or project. I suggest `scalapb.github.io/protobuf/scalapb`

protobuf/scalapb/scalapb.proto
```diff
@@ -2,7 +2,10 @@ syntax = "proto2";
 
 package scalapb;
 
+option csharp_namespace = "ScalaPb.Options";
```
thesamet, line 5: protoc accepts this file without this change, so this modification may break compatibility for users. Lets remove this line.

protobuf/scalapb/scalapb.proto
```diff
@@ -2,7 +2,10 @@ syntax = "proto2";
...
+option go_package = "scalapb.github.io/pirum.com/cda/grpc/trades";
 option java_package = "scalapb.options";
+option php_namespace = "ScalaPb\\Options";
```
thesamet, line 8: Currently, php generator only accepts proto files with proto3 syntax, so this line will not fix support for php users either. If php support is needed, let's open a different github issue to explore.

protobuf/scalapb/scalapb.proto
regiskuckaertz: woops haha

regiskuckaertz:
Thanks for your review, I have applied all changes.

protobuf/scalapb/scalapb.proto
```diff
@@ -2,6 +2,7 @@ syntax = "proto2";
 
 package scalapb;
 
+option go_package = "scalapb.github.io/cda/grpc/trades";
```
thesamet, line 5: There are still project specific bits here (cda, trades), can you replace the entire string with the one provided?

protobuf/scalapb/scalapb.proto
regiskuckaertz: Apologies, I was distracted when I read your comment.

protobuf/scalapb/scalapb.proto
regiskuckaertz: It's been a long day 😞 Thanks for your patience.

PR #1351, opened 2022-04-13
adding stub factory for clients

nadavwe:
This adds a stub factory to the client API.
I did it quick and dirty just to get the point across - I would gladly make changes.

I did not add tests since I wasn't sure how and where to add them - advice on that would be most welcome.

I decided to add an object but this can also be in a companion object or just on the main definition of the Grpc object.

Thanks for considering!

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
@@ -286,7 +286,9 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
       .newline
       .add(s"object ${service.name}StubFactory extends $stubFactory[${service.stub}] { ")
       .indent
-      .add(s"override def newStub(channel: $channel, options: $callOptions) = new ${service.stub}(channel, options)")
+      .add(
+        s"override def newStub(channel: $channel, options: $callOptions) = new ${service.stub}(channel, options)"
+      )
       .outdent
       .add("}")
       .newline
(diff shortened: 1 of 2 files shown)


thesamet:
Thanks for sending. I think that the stub factory should be the companion object of the client. BTW, what's the context that makes this change necessary?

nadavwe:
Hey @thesamet,
thanks for the review, I'll make the change soon.
Regarding tests again - where would you place them?

> BTW, what's the context that makes this change necessary?

I have a piece of code that has a delayed stub construction:
```scala
class BuldingStubThatNeedsMoreData[S <: AbstractStub[S]](factory: StubFactory[S]) {
  def buildWithExtraInfo(info: Info): S = {
    // use the info to build the channel and options
   factory.newStub(channel, options)
  }
}
```

the above call can now be called as: `new BuldingStubThatNeedsMoreData(new SomeStub(_, _))`.
But wouldn't it be nicer to have: `new BuldingStubThatNeedsMoreData(SomeStub)` like you suggest?
(or even `new BuldingStubThatNeedsMoreData[SomeStub]` with the correct type declarations naturally...)

thesamet:
Rational makes sense. The tests should go into the client section in `e2e-grpc/src/test/scala/GrpcServiceScalaServerSpec.scala`. In terms of what to test, it would be sufficient to test that the factory interface is generated and returns an instance of the correct type.

nadavwe:
@thesamet I think it's ok by me... please comment/review/merge.

nadavwe:
Thanks for the merge! happy Passover!

thesamet:
Thank you, and to you too! FYI, we have automatic snapshot releases, so this change is available here for testing https://oss.sonatype.org/content/repositories/snapshots/com/thesamet/scalapb/compilerplugin_2.12/0.11.10+11-32cd270e-SNAPSHOT/ . Let me know if you need a stable release. I am trying to batch a few changes between releases.

PR #1352, opened 2022-04-13
Make NonEmpty trait of sealed oneof universal

ya-goodfella:
So it may be mixed even with AnyVal types.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,2 @@
+-J-Xmx2G
+-J-XX:+UseG1GC
--- a/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
@@ -558,12 +558,12 @@ class DescriptorImplicits private[compiler] (
       deprecatedAnnotation ++ message.messageOptions.getCompanionAnnotationsList().asScala
     }
 
-    def extendsOption = messageOptions.getExtendsList.asScala.filterNot(valueClassNames).toSeq
+    def extendsOption = messageOptions.getExtendsList.asScala.filterNot(ValueClassNames).toSeq
 
     def companionExtendsOption = messageOptions.getCompanionExtendsList.asScala.toSeq
 
     def sealedOneofExtendsOption =
-      messageOptions.getSealedOneofExtendsList.asScala.filterNot(universalTraitNames).toSeq
+      messageOptions.getSealedOneofExtendsList.asScala.filterNot(UniversalTraitNames).toSeq
 
     def sealedOneOfExtendsCount = messageOptions.getSealedOneofExtendsCount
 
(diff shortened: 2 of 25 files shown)


thesamet:
Looks like this broke one of the e2e tests. Can you take a look?

ya-goodfella:
@thesamet I've introduced a new option called `sealed_oneof_universal_trait` for compatibility reasons.

Could you please check out the changes to code and documentation? It passes the tests locally now.

thesamet (commented):
Can this be accomplished by adding a list of base classes instead of a special option?

.sbtopts
```diff
@@ -0,0 +1,2 @@
+-J-Xmx2G
```
ya-goodfella, line 1: @thesamet 

Was getting `OutOfMemoryError` without `-Xmx` on my computer and even on Action runner. Added this for the project, if you don't mind.

ya-goodfella:
@thesamet looks like it's ready to be shipped. What do you think?

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -595,16 +596,20 @@ class DescriptorImplicits private[compiler] (
     def sealedOneofTypeMapper =
       sealedOneofTraitScalaType / (sealedOneofTraitScalaType.name + "TypeMapper")
 
-    private[this] val valueClassNames = Set("AnyVal", "scala.AnyVal", "_root_.scala.AnyVal")
+    private[this] val valueClassNames     = Set("AnyVal", "scala.AnyVal", "_root_.scala.AnyVal")
```
thesamet, line 599: Please move to `object DescriptorImplicits` below - there are a bunch of other constants there; first letter of the name Capitalized.

thesamet, line 599: Please move to `object DescriptorImplicits` below - there are a bunch of other constants there; first letter of the name Capitalized.

compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala
```diff
@@ -675,7 +680,8 @@ class DescriptorImplicits private[compiler] (
 
     def sealedOneofBaseClasses: Seq[String] = sealedOneofStyle match {
       case SealedOneofStyle.Default =>
-        messageOptions.getSealedOneofExtendsList.asScala.toSeq :+ "scalapb.GeneratedSealedOneof"
+        val any = if (isUniversalTrait) Seq("Any") else Seq.empty
```
thesamet, line 683: `maybeAny`

thesamet, line 683: `maybeAny`

e2e/src/main/protobuf/sealed_oneof_extends_universal.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+package com.thesamet.proto.e2e.universal;
```
thesamet, line 3: What would be a good test that the generator does the right thing and it's not just happening to compile? This would be useful to prevent regressions if changes in the generator are introduced later.

thesamet, line 3: What would be a good test that the generator does the right thing and it's not just happening to compile? This would be useful to prevent regressions if changes in the generator are introduced later.

ya-goodfella:
@thesamet kind reminder :)

thesamet:
Thank you!

Issue #1177, opened 2021-06-08
Unable to match on a repeated option field in a transformation


gontard:
I am trying to write the following rule based type customization:

```proto
message SetPermission {
   required string permission = 1 [(validate.rules).string = { in: ["manage", "view"] }]
}

option (scalapb.options) = {
  scope: PACKAGE
    {
      when: {
        options: {
          [validate.rules] {
            string {
              in: ["manage", "view"]
            }
          }
        }
      }
      set: {
        [scalapb.field] {
          type: "mypackage.Permission"
        }
      }
    }
  ]
};

```

but [this exception](https://github.com/protocolbuffers/protobuf/blob/4644980d818ec9d987f74c74b1a1d45de9ab01fd/java/core/src/main/java/com/google/protobuf/FieldSet.java#L1037-L1040) is raised during [transformation](https://github.com/scalapb/ScalaPB/blob/e368415a09a7d568ac81d8a13969fd3c05ef724f/compiler-plugin/src/main/scala/scalapb/compiler/FieldTransformations.scala#L98):

```
[info] Compiling 63 protobuf files to /home/gontard/stuff/target/scala-2.13/akka-grpc/main
--jvm_1_out: java.lang.IllegalArgumentException: hasField() can only be called on non-repeated fields.
	at com.google.protobuf.FieldSet.hasField(FieldSet.java:253)
	at com.google.protobuf.DynamicMessage.hasField(DynamicMessage.java:189)
	at scalapb.compiler.FieldTransformations$.$anonfun$matchContains$1(FieldTransformations.scala:98)
	at scalapb.compiler.FieldTransformations$.$anonfun$matchContains$1$adapted(FieldTransformations.scala:95)
  | => aat scala.collection.immutable.Map$Map1.forall(Map.scala:196)
	at scalapb.compiler.FieldTransformations$.matchContains(FieldTransformations.scala:95)
	at scalapb.compiler.FieldTransformations$.$anonfun$matchContains$1(FieldTransformations.scala:103)
	at scalapb.compiler.FieldTransformations$.$anonfun$matchContains$1$adapted(FieldTransformations.scala:95)

```

Am i doing something wrong or is it a limitation?

thesamet:
It's a bug/limitation - we haven't defined how matching should behave for repeated fields. I wonder what the behavior should be, possibly these two conditions should be met:
1. lists are of the same size
2. if elements are scalar, lists need to be equal
[...]

PR #1355, opened 2022-04-18
Support transformation on repeated option scalar

gontard:
Solves partially https://github.com/scalapb/ScalaPB/issues/1177 because it only add supports for scalar and not messages.
However it's enough for supporting `protoc-gen-validate` options.

### DIFF
--- a/compiler-plugin/src/test/scala/scalapb/compiler/FieldTransformationsSpec.scala
+++ b/compiler-plugin/src/test/scala/scalapb/compiler/FieldTransformationsSpec.scala
@@ -37,6 +37,7 @@ class FieldTransformationsSpec extends AnyFlatSpec with Matchers with ProtocInvo
          |}
          |message StringRules {
          |  optional string const = 1;
+         |  repeated string in = 2;
          |}
          |extend google.protobuf.FieldOptions {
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/test/scala/scalapb/compiler/FieldTransformationsSpec.scala
```diff
@@ -512,6 +533,26 @@ class FieldTransformationsSpec extends AnyFlatSpec with Matchers with ProtocInvo
       msg = "int32: {gt: 0, lt: 0}",
       pattern = "int32: {gt: 0}"
     ) must be(true)
+
```
thesamet, line 536: Can you add to docs/transformations.md a sentence describibing the semantics of matching repeated fields? I.e. empty vs non-empty is compared for presence. 
The behavior for CONTAINS can be a bit surprising, because it compares the lists for equality, not for containment. The behavior make sense, but let's document it.

The code looks good.

thesamet:
Thank you, Sébastien!

gontard:
Thanks for the quick turnaround.

Feel free to close #1177 (this fix is enough for me).

PR #1371, opened 2022-05-18
sealed_oneof_companion_extends option

ya-goodfella:


protobuf/scalapb/scalapb.proto
```diff
@@ -222,6 +222,9 @@ message MessageOptions {
   // file-level value, and can be overridden by the field-level setting.
   optional bool no_default_values_in_constructor = 9;
 
+  // Additional classes and traits to mix in to generated sealed_oneof base trait companion object.
+  repeated string sealed_oneof_companion_extends = 11;
```
ya-goodfella, line 226: index `10` already taken by #1352

thesamet:
Looks good. Please add a validation in `ProtoValidation.spec / validateMessage()` and a test in `ProtoValidationSpec` for the case where `sealed_oneof_companion_extends` is non-empty for a non-sealed one of message. There should be a `GeneratorException("selaed_oneof_companion_extends was provided for a non-sealed oneof message")`.

protobuf/scalapb/scalapb.proto
ya-goodfella: Nevermind, #1352 lost it's option. Changed back to `10` :)

.scalafmt.conf
```diff
@@ -2,10 +2,36 @@ version = 3.5.3
...
 assumeStandardLibraryStripMargin = true
-runner.dialect = "scala213source3"
+runner.dialect = scala213source3
 
 project.excludeFilters = [
    "scalapb-runtime/src/main/js-native/com/google/protobuf/.*",
    "scalapb-runtime/.*/scala.*/com/google/protobuf/.*",
    "scalapb-runtime/.*/scala.*/scalapb/options/.*"
 ]
+fileOverride {
```
thesamet, line 12: is this needed? generated files are not subject to scalafmt.

compiler-plugin/src/test/scala/scalapb/compiler/ProtoValidationSpec.scala
```diff
@@ -305,6 +305,23 @@ class ProtoValidationSpec extends AnyFlatSpec with Matchers with ProtocInvocatio
     )
   }
```
thesamet, line 307: Please add another test for sealed_oneofs_extends (looks like there's just one for companion)

.scalafmt.conf
ya-goodfella: Metals requested to update .scalafmt.conf accordingly. Maybe it may be ignored, but seems to me this will not do any harm.

If you're against - we may rollback it

compiler-plugin/src/test/scala/scalapb/compiler/ProtoValidationSpec.scala
ya-goodfella: nope, test for `sealed_oneof_extends` right above this one, on line 291

https://github.com/scalapb/ScalaPB/pull/1371/files/68a353f9de969743c9291b1b510065e3e52f7ae2#diff-a5d18260251210f14244e7f5ef58a497bca2972a74e5e4211df929ffde0b5ec4L291-L306

compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
```diff
@@ -103,10 +103,17 @@ class ProtoValidation(implicits: DescriptorImplicits) {
...
         )
       }
-    } else if (m.sealedOneOfExtendsCount > 0) {
-      throw new GeneratorException(
-        s"${m.getFullName}: is not a Sealed oneof and may not contain a sealed_oneof_extends message option. Use extends instead."
-      )
+    } else {
+      if (m.sealedOneOfExtendsCount > 0) {
+        throw new GeneratorException(
+          s"${m.getFullName}: is not a Sealed oneof and may not contain a sealed_oneof_extends message option. Use extends instead."
```
thesamet, line 109: s"${m.getFullName}: sealed_one_of_extend can only be used on sealed one ofs."   - (I think that the advice to use `extends` is unlikely to be relevant, it is more likely that the user misplaced the option, or assumes that something is a sealed signoff and it isn't.

ya-goodfella: Hm. It's double-sided, I suppose. They may just missclick on the IDE suggest, for example. :)

Maybe we may elaborate better help message in that case, pointing out that maybe user wanted to use a `oneof sealed_value` and a link to the documentation section over its use.

.scalafmt.conf
thesamet: My preference is to keep the PRs focused only on the changes necessary for the issue, and send changes like this as a separate PR. It would be great if you can roll back this file.

In the mean time how can I reproduce metals adding this locally?

ya-goodfella: I've just run "Import build" in VSCode

изображение

ya-goodfella: Nevermind, I've chosen "don't show again" xD

compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
```diff
@@ -103,10 +103,17 @@ class ProtoValidation(implicits: DescriptorImplicits) {
...
         )
       }
-    } else if (m.sealedOneOfExtendsCount > 0) {
-      throw new GeneratorException(
-        s"${m.getFullName}: is not a Sealed oneof and may not contain a sealed_oneof_extends message option. Use extends instead."
-      )
+    } else {
+      if (m.sealedOneOfExtendsCount > 0) {
+        throw new GeneratorException(
+          s"${m.getFullName} uses `sealed_oneof_extends` option while not being a sealed oneof (see https://scalapb.github.io/docs/sealed-oneofs). Did you mean `extends`?"
```
ya-goodfella, line 109: @thesamet what do you think about this phrasing?

compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
thesamet: > Hm. It's double-sided, I suppose. They may just missclick on the IDE suggest, for example.

I think that since it's ambiguous why the user did it a error message that elaborates on the two options may be too long and confusing. I think that a one line `"sealed_one_of_extend can only be used on sealed one ofs"` would set the user in the right direction without making assumptions.

compiler-plugin/src/main/scala/scalapb/compiler/ProtoValidation.scala
thesamet: I'd prefer: `${m.fullName}: sealed_one_of_extend can only be used on sealed one ofs. Did you mean `extends`?".

ya-goodfella: Done

ya-goodfella:
@thesamet do we need to fix anything else?

thesamet:
Thank you!

PR #1395, opened 2022-08-04
remove sbt-dotty plugin in docs

xuwei-k:


thesamet:
For Scala 3, the only thing users need to do is to set the scala version, so I think the entire page isn't useful any more.

PR #1396, opened 2022-08-06
Build for Scala Native, Scala 3

jodersky:
Since 0.4.4, Scala Native supports Scala 3.

The update to the munit testing library was necessary because the 0.x series does not support Scala Native on 3 ( see for a discussion https://github.com/scalameta/munit/pull/511).


thesamet:
Thank you!

PR #1433, opened 2022-11-11
Build scalapbc for arm64

visortelle:
Motivation is simple: we use [AWS Graviton](https://aws.amazon.com/ec2/graviton/) arm64-based instances.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,8 +74,12 @@ jobs:
         include:
         - os: ubuntu-latest
           arch: linux-x86_64
+        - os: ubuntu-latest
+          arch: linux-arm64
         - os: macos-12
           arch: osx-x86_64
+        - os: macos-12
+          arch: osx-arm64
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,31 +26,31 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '12.x'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -74,8 +74,12 @@ jobs:
         include:
         - os: ubuntu-latest
           arch: linux-x86_64
+        - os: ubuntu-latest
+          arch: linux-arm64
         - os: macos-12
           arch: osx-x86_64
+        - os: macos-12
+          arch: osx-arm64
     runs-on: ${{matrix.os}}
```
thesamet, line 83: The `arch` variable doesn't seem to impact the actual architecture used to run here. It only impacts the filename of the generated artifact. Wouldn't this just build an x86_64 arch but name the file `arm64`? Since this change runs only on release, are you able to somehow verify that this is expected to work in some test environment before the PR gets merged?

visortelle: OK, I'll check and submit changes in two-three days.

visortelle:
@thesamet As I figured out, GitHub Actions doesn't support GitHub-hosted arm64 runners at this moment.

I spend quite a lot of time without success trying to build it using this [action](https://github.com/uraimo/run-on-arch-action) that runs the build within QEMU.

I set up a self-hosted GitHub Actions runner in AWS. It was quite easy.
After installing missing C-related libs, and changing of GraalVM version the release process works.

Screenshot 2022-11-12 at 3 06 47 PM

Screenshot 2022-11-12 at 3 03 05 PM

It's not a good idea to use self-hosted runners with public repositories, therefore I won't leave it as is.

It's probably a good idea to wait for GitHub to add arm64 runners.

PR #1454, opened 2022-12-16
Bump protobuf-java

regiskuckaertz:
I was trying to get the latest zio grpc work with the latest scalapb but it seems like the `protobuf-java` version is not in sync and it is throwing a `NoSuchMethod` exception somewhere as the `parseUnknownField` has only been recently introduced (I believe due to security issues)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.typesafe.tools.mima.core._
 import BuildHelper._
 import Dependencies._
 
-val protobufCompilerVersion = "3.19.6"
+val protobufCompilerVersion = "3.21.12"
 
 val MimaPreviousVersion = "0.11.0"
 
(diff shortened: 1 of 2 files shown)


thesamet:
Bumping protobuf-java to 3.21.x would warrant a 0.12.x release for ScalaPB including a bunch of other libs, which I'd like to take more time for to consider additional changes/improvements in that release.

I believe the issue you are seeing is local to your project. ScalaPB doesn't use `parseUknownField`. Can you check if by any chance you are generating Java code using `PB.protocVersion > 3.19.6`? ScalaPB 0.11.x which is used by zio-grpc 0.5.x and 0.6.x relies on protobuf-java 3.19.6.

regiskuckaertz:
Oooh, ok I think I know what happened. Thanks for the pointers, I will have a look!

PR #1456, opened 2022-12-22
Update to grpc to 1.51.1 and protobuf to 3.21.12

fsonntag:
Updating grpc-java and protobuf-java. 

Would be great if we can get a new release with the updated versions

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -4,8 +4,8 @@ import Keys._
 
 object Dependencies {
   object versions {
-    val grpc                 = "1.47.1"
-    val protobuf             = "3.19.6"
+    val grpc                 = "1.51.1"
+    val protobuf             = "3.21.12"
     val silencer             = "1.7.12"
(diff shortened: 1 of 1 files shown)


thesamet:
Please see #1454. For grpc-java, any specific reason you can't upgrade it within your project? Is there any feature you need from protobuf-java 3.21.x?

fsonntag:
Ah, I see, thanks for getting back. Actually, I'm more keen on the GRPC update in order to have a newer `grpc-netty-shaded`, due to this issue:  
https://github.com/grpc/grpc-java/issues/8751

I'm not 100% sure if it was resolved with the newer version though

thesamet:
It should be possible to bump up the version of `grpc-netty-shaded` as a dependency in your build without a new version of scalapb.

fsonntag:
Yep, it is, I tried it! Was just worrying about conflicts appearing during runtime, so having the newer version in scalapb seems cleaner. But I'll with it for now, thanks a lot, @thesamet

PR #1458, opened 2022-12-31
Update documentation url

xuwei-k:


thesamet:
Thanks for the fixes! It looks like docusaurus is pretty flexible about the trailing slashes, but the links it generates on its own don't have them (unless there's an `#anchor`), so I removed some of the slashes. Thanks again!

PR #1459, opened 2022-12-31
fix `make_plugin_proto.sh` and re-generate code

xuwei-k:


thesamet:
Thanks for finding this. Submitted via c3010def

PR #1468, opened 2023-01-15
index fields by name and number to improve performance

oridag:


scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -86,9 +92,9 @@ class Descriptor private[descriptors] (
 
   def name: String = asProto.getName
 
-  def findFieldByName(name: String): Option[FieldDescriptor] = fields.find(_.name == name)
+  def findFieldByName(name: String): Option[FieldDescriptor] = fieldsByNameMap.get(name)
```
thesamet, line 95: Rather than creating a new map, we can leverage the map that exists at the field descriptor. `findSymbol` needs to be make `private[scalapb]`, the something like this should work:
 
```
file.findSymbol(fullName + "." + name) match {
  case res: Some(_: FieldDescriptor) => res
  case _ => None
}
```

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -77,6 +77,12 @@ class Descriptor private[descriptors] (
       FieldDescriptor.buildFieldDescriptor(fd, index, this)
     }.toVector
 
+  lazy val fieldsByNumberMap: Map[Int, FieldDescriptor] =
```
thesamet, line 80: I wonder if this helps performance for your use case. In scalapb-json4s we are using this method to search within map entry (which have only two fields: key and value), or primitive wrappers (1 value field). 

Can you do a benchmark with this method reverted?

If I am missing something and this is used in larger messages, following protobuf-java which has been extensively optimized, an alternative to the map here could be another vector (or array) sorted by field number, and this method can do a binary search. https://github.com/protocolbuffers/protobuf/blob/7679378c7f20fb814d75ecd375e68fc32ffea9ba/java/core/src/main/java/com/google/protobuf/Descriptors.java#L809-L812 , Assuming your benchmark suggest this doesn't matter for now we can defer this part of the change.

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
oridag: This change is actually not required for my use case case. I can revert it, or implement it as you suggest

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
oridag: This is the actually the change that solves the O(n^2) complexity for json parsing (I only added the byName change for completeness, I can revert it if you like). `findFieldByNumber` is called in `messageReads` for every field in the message

thesamet: Ack, my bad. let's add `private` sorted arrays of the fd's, and introduce a private `@tailrec` annotated binary search. Add a unit tests to `ScalaDescriptorSpec` leveraging existing messages to check the binary search for edge cases: empty message, message with 1 field like a primitive wrapper, and a large one like `TestAllTypes` from `unittest.proto`.

oridag: Why do you think a sorted array is better than a map?

oridag:
@nadavsr - changed "by name" look up as you suggested.
I also added a change in the code generation as an alternative way of fixing the performance issue. Let me know if you prefer that

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
thesamet: `Array[]` takes the least amount of memory possible; I anticipte lookup time to not vary match given the small size. I know Google's code is heavily benchmarked, so I'd opt for the same optimization they did if we're at it.

oridag:
@thesamet please review.
findByNumber uses binary search as you suggested.
findByName uses a local map because it turns out `FileDescriptor.descriptorsByName` does not contain field descriptors. I tried to add them but that's not currently possible without introducing mutability because `findSymbol` is called during field descriptors creation (which uses `descriptorsByName`, which is not yet initialized at that point).
Since field descriptors are not present on the file symbol map, I think its ok to have that map locally in the message descriptor (roughly the same memory footprint)

thesamet:
Thank you! You're right, getting the fields into the map used by `findSymbol` would introduce a tricky cycle.

It's probably a premature optimization, though I wanted to borrow a page from the official implementation and cut down on the number of maps, so I moved the map to the file-level, so at least we have one extra map per file rather than per message.

PR #1483, opened 2023-02-07
use eta-expansion in generated code if possible

xuwei-k:


compiler-plugin/src/main/scala/scalapb/compiler/ExpressionBuilder.scala
```diff
@@ -106,8 +118,12 @@ object ExpressionBuilder {
       convertCollection(s"""$e.map(__e => ${runSingleton(nontrivial)("__e")})""", targetType)
     else if (nontrivial.nonEmpty) {
       val f = nontrivial match {
-        case List(FunctionApplication(name)) =>
-          s"${name}(_)"
+        case List(ap: FunctionApplication) =>
+          if (ap.etaExpansion) {
```
thesamet, line 122: What are the cases when eta expansion is not ok?

xuwei-k: I’m not sure not ok cases in scalapb 🤔
In general, There are some cases can't use eta expansion in Scala.
e.g. the method have default params, type conversions. 

```
Welcome to Scala 2.13.10 (OpenJDK 64-Bit Server VM, Java 11.0.18).
Type in expressions for evaluation. Or try :help.

scala> def f1(x: Int, y: Int = 1) = ()
def f1(x: Int, y: Int): Unit

scala> List(2).map(f1)
                   ^
       error: type mismatch;
        found   : (Int, Int) => Unit
        required: Int => ?

scala> List(2).map(f1(_))
val res1: List[Unit] = List(())

scala> def f2(x: Int) = x
def f2(x: Int): Int

scala> List[Byte](3).map(f2)
                         ^
       error: type mismatch;
        found   : Int => Int
        required: Byte => ?

scala> List[Byte](3).map(f2(_))
val res3: List[Int] = List(3)


```

thesamet: Not sure if I'm misreading, but it looks like both these examples show that generally eta expansion is desirable. 

So reading this, I am still not sure what problem this PR is trying to solve. Can you file an issue that shows a possible use case that this PR solves?

thesamet: Let me know if there's still an interest in getting the PR in - trying to understand what type of problem this PR is solving.

PR #1503, opened 2023-04-05
MethodDescriptor: add input and output types

jordancoll:
what it says on the tin.

on gitter, there was mention that these fields would maybe be better exposed as `ScalaType` rather than the message descriptor directly. while I don't think this makes a huge difference, I think it makes consuming the endpoint a little more cumbersome. iiuc protobuf method in/out types will always be messages. upcasting to the more general `ScalaType` means calling code will just have to cast back down to `ScalaType.Message` to finally get at the underlying `Descriptor`.

I think a better alternative would be to return the `Descriptor` directly, since afaict wrapping it doesn't add value.
Alternatively, the fields could be typed as `ScalaType.Message` to avoid the up & down casts.

### DIFF
--- a/scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
+++ b/scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
@@ -149,34 +149,30 @@ class MethodDescriptor private[descriptors] (
 
   def getOptions = asProto.getOptions
 
-  lazy val inputType: ScalaType = {
+  lazy val inputType: Descriptor = {
     val inputType = asProto.inputType.getOrElse(
       throw new DescriptorValidationException(this, "missing method input type")
     )
 
-    ScalaType.Message(
-      FileDescriptor
-        .find(containingService.file, this, inputType)
-        .getOrElse(
-          throw new DescriptorValidationException(this, "couldn't build input type descriptor")
-        )
-        .asInstanceOf[Descriptor]
-    )
+    FileDescriptor
+      .find(containingService.file, this, inputType)
+      .getOrElse(
+        throw new DescriptorValidationException(this, "couldn't build input type descriptor")
+      )
+      .asInstanceOf[Descriptor]
   }
 
-  lazy val outputType: ScalaType = {
+  lazy val outputType: Descriptor = {
     val outputType = asProto.outputType.getOrElse(
       throw new DescriptorValidationException(this, "missing method output type")
     )
 
-    ScalaType.Message(
-      FileDescriptor
-        .find(containingService.file, this, outputType)
-        .getOrElse(
-          throw new DescriptorValidationException(this, "couldn't build output type descriptor")
-        )
-        .asInstanceOf[Descriptor]
-    )
+    FileDescriptor
+      .find(containingService.file, this, outputType)
+      .getOrElse(
+        throw new DescriptorValidationException(this, "couldn't build output type descriptor")
+      )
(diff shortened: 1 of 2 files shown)


thesamet:
I agree - it would be better to return the Descriptor.

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -148,6 +148,32 @@ class MethodDescriptor private[descriptors] (
   def location = containingService.file.findLocationByPath(SourceCodePath.get(this))
 
   def getOptions = asProto.getOptions
+
+  lazy val inputType: Descriptor = {
+    val inputType = asProto.inputType.getOrElse(
+      throw new DescriptorValidationException(this, "missing method input type")
```
thesamet, line 154: s"Missing method input type for method $fullName"

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -149,34 +149,30 @@ class MethodDescriptor private[descriptors] (
...
-      FileDescriptor
-        .find(containingService.file, this, inputType)
-        .getOrElse(
-          throw new DescriptorValidationException(this, "couldn't build input type descriptor")
-        )
-        .asInstanceOf[Descriptor]
-    )
+    FileDescriptor
+      .find(containingService.file, this, inputType)
+      .getOrElse(
```
thesamet, line 159: lets use pattern matching on `case Some(d: Descriptor)` to make the asInstanceOf unnecessary and handle the case that what's being found isn't a descriptor.

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -149,34 +149,30 @@ class MethodDescriptor private[descriptors] (
...
+        throw new DescriptorValidationException(this, "couldn't build input type descriptor")
```
thesamet, line 160: `Couldn't` ... `for $fullName`

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -149,34 +149,30 @@ class MethodDescriptor private[descriptors] (
...
+      )
+      .asInstanceOf[Descriptor]
   }
 
-  lazy val outputType: ScalaType = {
+  lazy val outputType: Descriptor = {
     val outputType = asProto.outputType.getOrElse(
       throw new DescriptorValidationException(this, "missing method output type")
```
thesamet, line 167: `Missing`

scalapb-runtime/src/main/scala/scalapb/descriptors/Descriptor.scala
```diff
@@ -149,34 +149,30 @@ class MethodDescriptor private[descriptors] (
...
-    ScalaType.Message(
-      FileDescriptor
-        .find(containingService.file, this, outputType)
-        .getOrElse(
-          throw new DescriptorValidationException(this, "couldn't build output type descriptor")
-        )
-        .asInstanceOf[Descriptor]
-    )
+    FileDescriptor
+      .find(containingService.file, this, outputType)
```
thesamet, line 171: Same as earlier comments (capital letter, avoid asInstanceOf through pattern matching)

PR #1525, opened 2023-06-06
Avoid warnings about unused values

som-snytt:
Noticed at https://discord.com/channels/632150470000902164/632150470000902166/1115280559023861800

This commit experiments with turning on lints, which helps test the following lintable changes.

The case is explicitly `()` to avoid inferring a LUB of `Any` or `AnyVal` for the match, since that value is unused. The actual LUB is `Unit`, which naturally never warns.

Two expressions are annotated `nowarn` because they are not consumed.

~One change is for builder's mutating method to return `this.type`, which means it doesn't quality as an unconsumed value. `skipField` from Java API is ascribed `: Unit` to fix the inferred type of the match (which was Any or AnyVal) and also indicate that discarding the value was intentional (though Java results don't qualify as unconsumed).~

Scala 3 doesn't like using a companion as a function unless it extends function. It is also safe to be explicit in Scala 2.

Two lints are turned off in test because of ScalaTest style.

~MiMA exclusions are added for the `this.type` change, but that could be handled also as `: Unit @nowarn` at use site.~

### DIFF
--- a/scalapb-runtime/src/main/scala/scalapb/UnknownFieldSet.scala
+++ b/scalapb-runtime/src/main/scala/scalapb/UnknownFieldSet.scala
@@ -89,7 +89,7 @@ object UnknownFieldSet {
           lengthDelimited = lengthDelimited.result()
         )
 
-      def parseField(tag: Int, input: CodedInputStream) = {
+      def parseField(tag: Int, input: CodedInputStream): this.type = {
         val wireType = WireType.getTagWireType(tag)
 
         wireType match {
@@ -106,6 +106,7 @@ object UnknownFieldSet {
               s"Protocol message tag had invalid wire type: ${wireType}"
             )
         }
+        this
       }
     }
 
@@ -135,9 +136,10 @@ object UnknownFieldSet {
       if (fieldBuilders.isEmpty) UnknownFieldSet.empty
       else new UnknownFieldSet(fieldBuilders.view.mapValues(_.result()).toMap)
 
-    def parseField(tag: Int, input: CodedInputStream) = {
+    def parseField(tag: Int, input: CodedInputStream): this.type = {
       val fieldNumber = WireType.getTagFieldNumber(tag)
       fieldBuilders.getOrElseUpdate(fieldNumber, new Field.Builder()).parseField(tag, input)
+      this
     }
   }
 }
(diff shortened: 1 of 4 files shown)


som-snytt:
Usually, I wait for test feedback before undrafting. Especially because I don't entirely know what I'm doing here.

scalapb-runtime/src/main/scala/scalapb/UnknownFieldSet.scala
```diff
@@ -89,7 +89,7 @@ object UnknownFieldSet {
           lengthDelimited = lengthDelimited.result()
         )
 
-      def parseField(tag: Int, input: CodedInputStream) = {
+      def parseField(tag: Int, input: CodedInputStream): this.type = {
```
thesamet, line 92: Curious, why `this.type` and not `Builder`?

som-snytt: The "unused warnings" don't warn in this case because `this.type` is not a value that is produced.

I made this change first (because it is a usual change; StringBuilder got this.type returns not long ago) but after making the other edit, maybe ignoring the value and adding nowarn is more general and doesn't mess with signatures.

som-snytt: And now I see it doesn't work on 2.12 anyway, which does have nowarn support IIRC. (I did run 2_13 e2e; oh wait, I thought warnings weren't on for 2.12?)

thesamet: If we use `this.type` to achieve what `nowarn` does, then I think we should just use `@nowarn` - generally keep things in simplest form and minimizing change. I'm ok with turning off warnings for 2.12 at this stage...

som-snytt: OK I'll update with nowarn, thanks.

thesamet:
Hi @som-snytt , checking in if you'd like to continue working on this PR?

som-snytt:
Thanks, I've been reducing my personal backlog and this was still on it. I refreshed my unreliable memory and improved the tweaks. I think I didn't get the warnings right on 2.12, so that will require a fix. I can squash when it looks OK.

Issue #1316, opened 2022-01-31
Support new Scala 3 syntax


Lasering:
scalapb generated code fails to compile in Scala 3 under these scalac flags:

- `-new-syntax` - Require `then` and `do` in control expressions.
- `-source:future` - Deprecates: private[this], `_` for wildcard arguments, some other things
- `-Werror` - Fail the compilation if there are any warnings.

Would it be possible to have a new code generator under Scala 3 which would fix these problems (some of them will have to be fixed sooner or later)? I can make a PR if it helps (I'll probably need a little guidance.)

thesamet:
I assume the reason this change is needed since you have a single project that includes both your code and generated code (correct me if I'm wrong). A possible solution which I highly recommend, regardless of this problem, is to move the code generation in a dedicated sub-project within the same SBT build, and have the main project depend on it. Currently, the generator aims to generate code that works for both Scala 2.x and 3.x. This reduces the effort needed to keep the code updated. I would defer on satisfying the above flags until they are required by default by scalac.

Lasering:
You are correct. Another sub-project would solve it.

According to [Wildcard Arguments in Types](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html):
> In Scala 3.2, the meaning of _ changes from wildcard to placeholder for type parameter.

We are already at Scala 3.1.1 so it won't take long (assuming they stick to that schedule) for these changes to be necessary.

thesamet:
Yes, maybe having the code generator have cases for different language versions is going to be inevitable one day. Is there an instance where the generated code is impacted by wildcard arguments in types?

Lasering:
Using the [Address Book example from the docs](https://scalapb.github.io/docs/getting-started#defining-your-protocol-format) and the `-source:future`, `-Werror` flags the following errors are reported regarding the wildcard arguments:

```
[error] -- Error: (...)/modules/shared/.jvm/target/scala-3.1.1/src_managed/main/scalapb/tutorial/addressbook/AddressBook.scala:98:107 
[error] 98 |  def messageCompanionForFieldNumber(__number: _root_.scala.Int): _root_.scalapb.GeneratedMessageCompanion[_] = {
[error]    |                                                                                                           ^
[error]    |      `_` is deprecated for wildcard arguments of types: use `?` instead
[...]

PR #1565, opened 2023-09-04
support new scala 3 syntax

ngbinh:
This PR adds a new gen option for scala 3. It should generate code without any warnings with source:future scalacOptions.

should address https://github.com/scalapb/ScalaPB/issues/1316


compiler-plugin/src/main/scala/scalapb/compiler/GeneratorParams.scala
```diff
@@ -49,6 +50,8 @@ object GeneratorParams {
             case "no_lenses" => Right((params.copy(lenses = false), unrecognized))
             case "retain_source_code_info" =>
               Right((params.copy(retainSourceCodeInfo = true), unrecognized))
+            case "scala3Sources" =>
```
thesamet, line 53: For consistency with the other flags, it should be "scala3_sources".

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -23,6 +23,22 @@ class ProtobufGenerator(
   import DescriptorImplicits.AsSymbolExtension
   import ProtobufGenerator._
 
+  private val CompSeqType = {
```
thesamet, line 26: I prefer that constants will go inside `object C {}` (I know it's currently inconsistent). Since this constant depends on the parameteres, can you change `object C` to `class Constants(params: ...)`, and inside `class ProtobufGenerator` have `private val C = new Constants(params)` so the change is minimized?

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1048,14 +1072,23 @@ class ProtobufGenerator(
   def generateMessageCompanionMatcher(methodName: String, messageNumbers: Seq[(Descriptor, Int)])(
       fp: FunctionalPrinter
   ): FunctionalPrinter = {
-    val signature =
+    val signature = if (params.scala3Sources) {
+      s"def $methodName(__number: _root_.scala.Int): _root_.scalapb.GeneratedMessageCompanion[?] = "
```
thesamet, line 1076: Since there are a bunch of `if..else` s around the wildcard, I suggest we add a `WildcardType` constant to `Constants`. Then here and below just interpolate using `WildcardType`.

lenses/src/main/scala-2.12/scalapb/lenses/Lenses.scala
thesamet, line 1: What's changing about lenses? Is it necessary?

scalapb-runtime/src/main/scala-2.13/scalapb/descriptors/PValue.scala
thesamet, line 1: It looks like there are two copies of `PValue`, can you explain why it's needed?

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -1249,15 +1285,24 @@ class ProtobufGenerator(
     }
   }
 
+  private val WithOperator = {
```
thesamet, line 1288: Should be move to `Constants`

thesamet:
Thanks! To silence the mima issues, please add an exclusion in build.sbt along the lines of:
```
ProblemFilters.exclude[DirectMissingMethodProblem]("scalapb.compiler.GeneratorParams.*")
```

build.sbt
```diff
@@ -28,8 +28,8 @@ lazy val root: Project =
     .in(file("."))
```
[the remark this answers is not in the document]
ngbinh: sure, this is probably because I tried to run `fmt`. Will double check

compiler-plugin/src/main/scala/scalapb/compiler/GeneratorParams.scala
ngbinh: good point.

lenses/src/main/scala-2.12/scalapb/lenses/Lenses.scala
ngbinh: `ObjectLense` class needed to be declared as [open class](https://docs.scala-lang.org/scala3/reference/other-new-features/open-classes.html). Right now, I can't find a better way for doing so only for scala-3 without duplicate the class across all scala flavors.

scalapb-runtime/src/main/scala-2.13/scalapb/descriptors/PValue.scala
ngbinh: For scala 3, we need

```scala
sealed trait PValue extends Any, Matchable {
  def as[A](implicit reads: Reads[A]): A = reads.read(this)
}
```

I have tried to declare the base type inside `scalapb.internal.compat` but I couldn't do so.

ngbinh:
@thesamet addressed all of your comments. Thanks

lenses/src/main/scala-2.12/scalapb/lenses/Lenses.scala
thesamet: Can't we add `import scala.language.adhocExtensions` in the source files to suppress this problem?

ngbinh: I will try to go down that route.

ngbinh:
update: bring back a single Lense and add in adhocExtension import for scala 3

ngbinh:
@thesamet please have another look when you have time. Thanks

thesamet:
Merged via https://github.com/scalapb/ScalaPB/commit/83b849f59764f527c2ecda4905e4e48d75bbb4b2 - thank you!

ngbinh:
Thanks @thesamet!

PR #1590, opened 2023-10-10
Add no_getters generator parameter

blaenk:
If I'm not mistaken, it seems like https://github.com/scalapb/ScalaPB/pull/1140 may have forgotten to expose a generator parameter like `no_lenses` but for getters.

Thank you for working on ScalaPB!


thesamet:
Hi @blaenk , I appreciate the time you took to put together the PR. The project preference is to have most customizations only available as package-scoped options. Please see https://scalapb.github.io/docs/faq/#why-a-certain-customization-is-not-available-as-a-global-generator-parameter

blaenk:
Ah that makes sense. Thankfully this PR was trivial enough to write so not much time was spent. Thanks for letting me know!

PR #1663, opened 2024-03-15
Update protobuf to 25.3

molekyla:
Hi. I've fount that ScapaPB relies on quiet old version of the Protobuf. Let's upgrade it unless there is no reasons to keep it

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import BuildHelper._
 import Dependencies._
 import sbtassembly.AssemblyPlugin.defaultUniversalScript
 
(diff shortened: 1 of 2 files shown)


thesamet:
Upgrading can impose binary incompatibilities on users, and generally if users need a new version they can do so locally in their project. I am looking into migrating to 4.26.0 - which require some change. Is there a reason why 25.3 was chosen?

molekyla:
The version 3.25.3 was chosen becaus I missed the release of 4.26.0. It would be great to have freshest version.
Is there any planned date when new version (with protobuf 4.26.0) will be released?

PR #1682, opened 2024-04-19
Update nscplugin, sbt-scala-native to 0.5.1

avdv:
- Update nscplugin, sbt-scala-native to 0.5.1
- Use `nativeConfig` to enable link stubs

Adapted from #1680

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,7 @@ object Dependencies {
     val annotationApi           = "1.3.2"
     val cats                    = "2.6.1"
     val mockito                 = "5.11.0"
-    val munit                   = "1.0.0-M8"
+    val munit                   = "1.0.0-M12"
     val scalaTest               = "3.2.18"
     val scalaTestPlusMockito    = "3.1.0.0"
     val scalaTestPlusScalaCheck = "3.2.18.0"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,7 +10,7 @@ object Dependencies {
     val collectionCompat     = "2.12.0"
     val coursier             = "2.1.9"
     val protocGen            = "0.9.7"
-    val protobufRuntimeScala = "0.8.14"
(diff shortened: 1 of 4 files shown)


thesamet:
It looks like scala-collection-compat need to be updated to latest version too.

avdv:
> It looks like scala-collection-compat need to be updated to latest version too.

Right. I pushed a commit to update to 2.12.0.

avdv:
Seems like munit needs to be updated too. I updated it to version 1.0.0-M11.

thesamet:
Can you upgrade the JVM to Java 11 for this specific test target only?

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
     val annotationApi           = "1.3.2"
     val cats                    = "2.6.1"
     val mockito                 = "5.11.0"
-    val munit                   = "1.0.0-M8"
+    val munit                   = "1.0.0-M11"
```
tgodzik, line 20:
```suggestion
    val munit                   = "1.0.0-M12"
```
we updated to newer version already, M11 might be broken.

avdv: :ok_hand: done.

avdv:
> Can you upgrade the JVM to Java 11 for this specific test target only?

:ok_hand: done.

Alternatively, how about downgrading mockito to 4.x?

avdv:
ping ...

thesamet:
Releasing protobuf-scala-runtime 0.8.15 which support native 0.5.

MateuszKubuszok:
Shouldn't `scalapb-runtime` be released for `native0.5` as well? Without it any code that uses:

```scala
libraryDependencies += "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
```

does not compile.

thesamet:
Please update to ScalaPB 0.11.17.

PR #1765, opened 2024-11-14
Add information about which required field was missing

gregor-i:
Exceptions caused by missing required fields don't mention which fields were missing, even though this information is easy to provide


thesamet:
Can you add unit tests? They can be based on e2e/src/main/protobuf/google/protobuf/unittest.proto and use `TestRequired`, `TestRequiredForeign`, and `TestRequiredOneof`, it should exercise the cases were one or more than one fields are missing.

gregor-i:
I've added some test cases. I didn't know how to construct incomplete messages, so my test data setup may be overcomplicated.

PR #1774, opened 2024-11-28
Pass `safe` and `idempotent` options to MethodDescriptor

igor-vovk:
Hi! I'm working on the library that provides ConnectRPC-compatible REST interface from ScalaPB service interfaces https://github.com/igor-vovk/connect-rpc-scala. 

One of the nice features of the Connect protocol is being able to expose endpoints so they are reachable with GET-requests instead of POST (https://connectrpc.com/docs/protocol/#unary-get-request). It is controlled by the `idempotency_level` option. This PR adds passing this option to the MethodDefinition, so that https://github.com/igor-vovk/connect-rpc-scala/pull/14/files#diff-c75d247c647e264575f9907d360db7a0769a44833b041a1a540b2eb7592ca656R71 would be possible.

### DIFF
--- a/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
+++ b/compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
@@ -178,11 +178,11 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
       case StreamType.Bidirectional   => "BIDI_STREAMING"
     }
 
-    val grpcMethodDescriptor = "_root_.io.grpc.MethodDescriptor"
-
     val idempotencyLevel = method.getOptions.getIdempotencyLevel
     val safe             = idempotencyLevel == IdempotencyLevel.NO_SIDE_EFFECTS
-    val idempotent       = idempotencyLevel == IdempotencyLevel.IDEMPOTENT
+    val idempotent       = idempotencyLevel == IdempotencyLevel.IDEMPOTENT || safe
+
+    val grpcMethodDescriptor = "_root_.io.grpc.MethodDescriptor"
 
     p.add(
       s"""${method.deprecatedAnnotation}val ${method.grpcDescriptor.nameSymbol}: $grpcMethodDescriptor[${method.inputType.scalaType}, ${method.outputType.scalaType}] =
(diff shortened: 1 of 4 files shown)


compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -179,12 +180,18 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
 
     val grpcMethodDescriptor = "_root_.io.grpc.MethodDescriptor"
 
+    val idempotencyLevel = method.getOptions.getIdempotencyLevel
+    val safe             = idempotencyLevel == IdempotencyLevel.NO_SIDE_EFFECTS
+    val idempotent       = idempotencyLevel == IdempotencyLevel.IDEMPOTENT
```
thesamet, line 185: No side effects implies idempotent. Should there be a logical or here? https://protobuf.dev/reference/java/api-docs/com/google/protobuf/DescriptorProtos.MethodOptions.IdempotencyLevel.html#NO_SIDE_EFFECTS_VALUE

compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
```diff
@@ -179,12 +180,18 @@ final class GrpcServicePrinter(service: ServiceDescriptor, implicits: Descriptor
...
+
     p.add(
       s"""${method.deprecatedAnnotation}val ${method.grpcDescriptor.nameSymbol}: $grpcMethodDescriptor[${method.inputType.scalaType}, ${method.outputType.scalaType}] =
          |  $grpcMethodDescriptor.newBuilder()
          |    .setType($grpcMethodDescriptor.MethodType.$methodType)
          |    .setFullMethodName($grpcMethodDescriptor.generateFullMethodName("${service.getFullName}", "${method.getName}"))
          |    .setSampledToLocalTracing(true)
+         |    .setSafe($safe)
```
thesamet, line 193: Can we `setSafe` and `setIdempotent` only when `method.getOptions.hasIdempotencyLevel` ? You cal use `p.when` to emit code conditionally.

compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
igor-vovk: Thanks, fixed

compiler-plugin/src/main/scala/scalapb/compiler/GrpcServicePrinter.scala
igor-vovk: Thanks for spotting it. Also found the tests and added a test for this case

igor-vovk:
@thesamet fixed your previous comments 🙏

igor-vovk:
I'm wondering if there is anything else that needs to be done in this PR so it will be merged

thesamet:
Sorry, this fell off my radar. Merged!

PR #1862, opened 2025-06-01
use ubuntu 22 in release.yml

pjfanning:
GitHub CI won't run ubuntu 20 any more.
Example: https://github.com/scalapb/ScalaPB/actions/runs/15352793795


thesamet:
Thanks for the PR. Upgraded to 24.04 via another commit.

Issue #1911, opened 2025-06-16
ScalaPB generates incorrect import paths for `oneof` fields when field name conflicts with imported package


ankita-ndd:
When using ScalaPB to generate Scala code from protobuf definitions, there's an issue with import path resolution when a oneof field name conflicts with an imported package name.

Would like to understand if there is a way to workaround this error without changing the proto definition, such as any options that can be passed in? I tried a few but did not get the results. 
Alternatively, let me know if this proto definition is a violation of naming standard in protocol buffers, as such I could not find a note indicating that but my experience with protocol buffers is relatively new. 

**Example:**
```
import "location/v1/location.proto";

message SomeRequest {
  oneof location {  // This field name conflicts with the imported package
    location.v1.Coordinate location_coordinate = 2;
  }
}
```
**Current Behavior:**
The generated Scala code incorrectly tries to access location.v1.location.Coordinate as if it were nested under the Location class:
```
value v1 is not a member of labor.v1.SomeRequest.Location
-- [E008] Not Found Error: <path-to>/v1/SomeRequest.scala:247:73 
247 |    def locationCoordinate: _root_.scalapb.lenses.Lens[UpperPB, location.v1.Coordinate] = field(_.getLocationCoordinate)((c_, f_) => c_.copy(location = labor.v1.SomeRequest.Location.LocationCoordinate(f_)))
```

**Expected Behavior:**
The generated code should correctly reference the imported type:
```
def getLocationCoordinate: location.v1.Coordinate = location.locationCoordinate.getOrElse(location.v1.Coordinate.defaultInstance)
```
**Impact:**
- This causes compilation errors in the generated Scala code
- Forces users to either rename their oneof fields or modify their proto files to work around the issue
- Affects code generation when using Scala 3

**Environment:**
ScalaPB version: https://repo1.maven.org/maven2/com/thesamet/scalapb/protoc-gen-scala/0.11.18/ 
Protoc version: libprotoc 29.3
Scala version: 3.6.4

i-am-the-slime:
I think I'll need some pointers how anything works here, or I will have to resort to ask the AI. Whichever you prefer 😁

thesamet:
Either way, though I'll be out for the next two weeks :)

ae0n:
Hi @thesamet if you have time can you please check PR I made to fix this issue

i-am-the-slime:
@thesamet it would really be great for us if we could move this thing forward

dmytr:
[...]

PR #1912, opened 2025-06-27
resolve oneof field name conflicts with imported packages

ae0n:
Include `oneof` field names in conflict detection to properly generate _root_ prefixes when `oneof` fields conflict with imported package names.
Closes https://github.com/scalapb/ScalaPB/issues/1911


i-am-the-slime:
@thesamet do you think we should fork the library?

PR #1947, opened 2025-08-22
Fix test class name for oneof import conflict

i-am-the-slime:
Fixes the compilation error in PR #1912 by correcting the test class name from `OneofImportConflict` to `OneofImportConflictTest`.

Closes #1912


i-am-the-slime:
Looks like the workflow needs approval. Maybe @nadavsr you could approve it?

Issue #1908, opened 2025-06-11
FieldOptions.fromJavaProto does not include unknownFields


silverbeak:
Hello.

**Quick note**: 
First-time poster here, and I've read the contribution guidelines, but I may still miss a few details. If so, apologies in advance, any and all guidance is welcome.

**Context**:
I've created a custom DynamicMessage type, similar to the one in google's library.
In order to instantiate this message type, I'm reading a serialized in the following fashion:
```scala
//** Serializing side **//
val descriptor = HelloMessage.scalaDescriptor.file
val serialized = com.google.protobuf.descriptor.FileDescriptorProto.toByteArray(descriptor.asProto)


//** De-serializing side **/
// The javaFileDescriptorProto contains the proper options/extensions
val javaFileDescriptorProto = com.google.protobuf.DescriptorProtos.FileDescriptorProto.parseFrom(serialized)

// But the scalaFileDescriptorProto does not contain the options/extensions...
val javaFileDescriptor = FileDescriptorProto.fromJavaProto(javaFileDescriptorProto)

// ...and subsequently, the resulting fileDescriptor is also missing the options/extensions
val fileDescriptor = FileDescriptor.buildFrom(javaFileDescriptor, Nil)
```

**Problem**
The code above will compile and run fine, but when I inspect the resulting object I'm missing data in the following location.

`fileDescriptor -> messages[msgIndex] -> fields[fieldIndex] -> asProto -> options`

In practice, this means that my decorator extensions will not be properly de-serialized.

**Solution?**
I noticed that unknownFields are not handled [here](https://github.com/scalapb/ScalaPB/blob/35baf0e7fbf26e7fc203772fce60ce31aeedb511/scalapb-runtime/src/main/scalajvm/com/google/protobuf/descriptor/FieldOptions.scala#L343). So I had a bit of a tinker and wrote a rough implementation of what it is I would like to achieve. The implementation can be seen [here](https://github.com/silverbeak/ScalaPB/blob/58c9e612fd58e164dbc5a1660db9836655a93f8a/scalapb-runtime/src/main/scalajvm/com/google/protobuf/descriptor/FieldOptions.scala#L358) and [here](https://github.com/silverbeak/ScalaPB/blob/58c9e612fd58e164dbc5a1660db9836655a93f8a/scalapb-runtime/src/main/scalajvm/com/google/protobuf/descriptor/FieldOptions.scala#L1239).

[...]

PR #1963, opened 2025-09-11
Adding unknown fields when parsing fieldoptions from javaproto

silverbeak:
I created an issue for this [here](https://github.com/scalapb/ScalaPB/issues/1908). It didn't get any attention, but I'll just go ahead and try a PR.

Please let me know if I need to make any changes. Thanks.


silverbeak:
I added a test for this. The test would fail before this PR, but now it runs successfully.
Please let me know if this is not what you had in mind.

thesamet:
AFK for the next 3 weeks. Will review when I am back

-Nadav

PR #2055, opened 2026-02-10
Added Drainable implementation to ProtoInputStream

laoqiu233:
The current default output stream that the scalapb marshaller uses does not support the [Drainable interface](https://github.com/grpc/grpc-java/blob/9d5842b8133c87e74b9300751710192231485662/api/src/main/java/io/grpc/Drainable.java#L31) from grpc-java, which becomse a problem [here](https://github.com/grpc/grpc-java/blob/9d5842b8133c87e74b9300751710192231485662/core/src/main/java/io/grpc/internal/MessageFramer.java#L279) where high performance code are expected to support it to avoid copying of the serialized protobuf.

This relatively simple fix allowed us to optimize a performance critical and high throughput service, and after deploying it to production we managed to cut a few GB of heap usage in the runtime and significantly lowered the heap presuure, but we did not get noticible benefit in timings. (Although for bigger protobuf messages it could make the timings a lot better by avoiding the copying)

### DIFF
--- a/scalapb-runtime-grpc/src/main/scala/scalapb/grpc/ProtoInputStream.scala
+++ b/scalapb-runtime-grpc/src/main/scala/scalapb/grpc/ProtoInputStream.scala
@@ -1,35 +1,39 @@
 package scalapb.grpc
 
 import com.google.protobuf.CodedOutputStream
+import io.grpc.Drainable
 import scalapb.GeneratedMessage
 
-import java.io.{ByteArrayInputStream, InputStream}
+import java.io.{ByteArrayInputStream, InputStream, OutputStream}
 
 /** Allows skipping serialization completely when the io.grpc.inprocess.InProcessTransport is used.
   * Inspired by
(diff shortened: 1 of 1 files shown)


scalapb-runtime-grpc/src/main/scala/scalapb/grpc/ProtoInputStream.scala
```diff
@@ -1,35 +1,39 @@
...
     def read(): Int
     def read(b: Array[Byte], off: Int, len: Int): Int
+    def drainTo(target: OutputStream): Int
   }
 
   private object Drained extends State {
     override def available: Int                                = 0
     override def read(): Int                                   = -1
     override def read(b: Array[Byte], off: Int, len: Int): Int = -1
+    override def drainTo(target: OutputStream): Int            = -1
```
thesamet, line 29: Look at https://grpc.github.io/grpc-java/javadoc/io/grpc/Drainable.html#drainTo(java.io.OutputStream), the return value should be number of bytes written. Shouldn't this return 0?

thesamet:
Is this change correct in the sense that after draining the state should change? Multiple calls to drainTo should leave the object drained (the second call should be a no-op).

PR #1376, opened 2022-05-23
[WIP] #1373 lazy fields

ya-goodfella:
A proof-of-concept implementation of #1373. Under construction, published as draft, beware the dragons.

PR #2072, opened 2026-03-08
Lazy fields parsing 

poslegm:
Hello!

This is a revival of the work to support lazy fields (previous attempt https://github.com/scalapb/ScalaPB/pull/1376).

#### Context
In Java protobuf, string fields are handled using [LazyField](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/LazyField.java). This mechanism stores the field data as a ByteString and only parses it into a UTF‑8 string when the corresponding getter is called. When a message containing such fields is serialized, the raw `ByteString` is written directly, without performing UTF‑8 encoding or decoding ([source](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/LazyFieldLite.java#L456-L466)).

Unlike java protobuf, scalapb does not lazily serialize strings. Accordingly, this is an opportunity to reduce the overhead if the following factors coincide:
- protobuf message consists of a large number of `string` fields;
- read access (calling a getter method) to only a small number of attributes (parse message → read a few fields → serialize).

Such usage patterns are quite common for cloud-native applications.

#### The essence of the changes

Generating `LazyField[String]` for string fields if `scalapb.options.lazy_fields` is enabled. `LazyField[T]` contains  the original `ByteString` and lazily parses the value on demand. Introduces implicit conversions for convenient use of generated case classes. 

Example:

```protobuf
message LazyWithRecursion {
  option (scalapb.message) = {
    lazy_fields: true
  };
  string data = 1;
  LazyWithRecursion nested = 2;
}
```

```scala
val original = LazyWithRecursion(data = "a lazy string", nested = Some(LazyWithRecursion(data = "nested string")))

val updated = original.update(_.nested.data := "updated string")

println(updated.nested.get.data == "updated string") // true
```

How it works with parsing and serialization:
```scala
val msg = LazyMessage.parseFrom(bytes)

// No parsing has happened for lazy_field yet.

val serialized = msg.toByteArray // <--- fast serialization without UTF-8 encoding

val upper = msg.lazyField.toUpperCase  // <--- Parsing happens here, on first access.

println(upper)
```

#### Benchmarks
New benchmarks have been added: `roundTripScala` and `roundTripJava`. They test the full proto lifecycle: parsing and serialization.  I was confused by the fact that transforming data in `object ${Message}Test` using the `toJavaProto` method affects the performance results. In my generated code, this method forces `ByteString` usage during java proto preparation, so the comparison is not entirely fair. **The results have also improved for existing benchmarks**, but I wanted a clearer comparison.

| Round-trip benchmark                                                       | Java        | Scala       |
| -------------------------------------------------------------------------- | ----------- | ----------- |
| LargeStringMessage                                                         | 9,345 ns/op | 9,088 ns/op |
| LazyFieldsStringMessage (same as LargeStringMessage but lazy_fields: true) | 9,484 ns/op | 2,734 ns/op |

Looks great. More than 3x speedup 🚀 Of course, scalapb is faster than java proto even without additional improvements.

#### Questions
1. Should I commit benchmarks results into this PR?
2. Should I add more tests?
3. Should we keep using `toJavaProto` in data preparation for benchmarks (`object ${Message}Test`)?
4. Anything else?


benchmarks/src/main/templates/message.scala.tmpl
poslegm: Potential problem with data preparation is here. Java protobuf created by conversion from scala proto instead of bytes parsing.

So, lazy_fields affects java serialization performance:

| serializeJava                                                      | Java        |
| -------------------------------------------------------------------------- | ----------- |
| LargeStringMessage                                                         | 2,584 ns/op |
| LazyFieldsStringMessage (same as LargeStringMessage but lazy_fields: true) | 1,111 ns/op |

i think that it is caused by forced bytes writing at `toJavaProto` with lazy_fields enabled (ProtobufGenerator#348L). But anyway it doesn't look as clear java protobuf benchmark.

What about changing this line to `val java = Protos.${Message}.parseFrom(bytes)`?

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
```diff
@@ -0,0 +1,54 @@
...
+  *   the decoder to use to parse the bytes.
+  * @tparam T
+  *   the type of the field.
+  */
+final class LazyField[T] private (val bytes: ByteString, decoder: LazyDecoder[T]) {
+  lazy val value: T = decoder.decode(bytes)
+
+  def toByteString: ByteString = bytes
+
+  override def toString: String = value.toString()
```
thesamet, line 45: `toString` forces lazy decoding. Any logging, string interpolation (`s"$msg"`), or debug print will silently trigger UTF-8 parsing — defeating the round-trip optimization. Consider `s"LazyField(${bytes.size()} bytes)"` here and letting callers use `.value.toString()` explicitly.

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
```diff
@@ -0,0 +1,54 @@
...
+  // Equality for LazyField[T] is not commutative!
+  // It is extremely important to use LazyField[T] only with the `-language:strictEquality` enabled for Scala 3 or `-Xfatal-warnings` for Scala 2.
+  override def equals(other: Any): Boolean = other match {
```
thesamet, line 48: The non-commutative `equals` is the most concerning aspect of this public API. The tests explicitly demonstrate that `Set[Any](s, lazyS)` and `Set[Any](lazyS, s)` have different sizes depending on insertion order. This is a silent correctness hazard for code using `Any`-typed collections or comparing against extracted strings. The comment recommends `-language:strictEquality`, but that's Scala 3 only and opt-in — Scala 2 users have no protection. Please document this limitation prominently in the scaladoc on the class.

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
```diff
@@ -0,0 +1,54 @@
...
+
+trait LazyDecoder[T] {
+  def decode(bytes: ByteString): T
+}
+
+object LazyDecoder {
+  implicit val stringDecoder: LazyDecoder[String] = _.toStringUtf8()
+}
+
+trait LazyEncoder[T] {
```
thesamet, line 77: `LazyEncoder` is defined but not used anywhere in the implementation (the write path goes through the TypeMapper's `toBase` which calls `.toByteString` directly). Either use it or remove it to avoid dead public API surface.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
```diff
@@ -419,9 +428,11 @@ class ProtobufGenerator(
       case FieldDescriptor.JavaType.DOUBLE      => FunctionApplication(s"$d.PDouble")
       case FieldDescriptor.JavaType.BOOLEAN     => FunctionApplication(s"$d.PBoolean")
       case FieldDescriptor.JavaType.BYTE_STRING => FunctionApplication(s"$d.PByteString")
-      case FieldDescriptor.JavaType.STRING      => FunctionApplication(s"$d.PString")
-      case FieldDescriptor.JavaType.ENUM        => FunctionApplication(s"$d.PEnum")
-      case FieldDescriptor.JavaType.MESSAGE     => MethodApplication("toPMessage")
+      case FieldDescriptor.JavaType.STRING if fd.getContainingType.lazyStringFields =>
```
thesamet: Mapping lazy string fields to `PByteString` in `singleFieldAsPvalue` breaks JSON serialization. Libraries like `scalapb-json4s` and `scalapb-circe` call `getField` (which uses this path) to convert messages to JSON. With this change, lazy string fields would be emitted as **base64-encoded bytes** instead of JSON strings.

The inverse direction is equally broken: `generateMessageReads` uses `baseSingleScalaTypeName = ByteString` for lazy fields, so it looks for `PByteString` in the field map. But JSON parsers produce `PString` for string fields — so JSON parsing of messages with lazy string fields would silently return default values.

Fix: keep the `PString` path for `singleFieldAsPvalue` (call `.value` on the lazy field to get the string before wrapping), and handle the `PString -> LazyField[String]` conversion in `generateMessageReads`.

e2e/src/test/scala/LazyStringFieldsSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  "Lazy repeated fields" should "work correctly through serialization" in {
+    val originalItems = Seq("hello", "world", "again")
+    val original = LazyRepeated(items = originalItems)
+    
+    val bytes = original.toByteArray
+    val parsed = LazyRepeated.parseFrom(bytes)
+
+    def f(str: Seq[String]): Int = str.length
```
thesamet, line 54: `parsed.items` has type `Seq[LazyField[String]]` (because `lazyStringFields` is enabled at file level), but `f` expects `Seq[String]`. There is no `Conversion[Seq[LazyField[String]], Seq[String]]` defined — implicit conversions don't apply element-wise to collections. This should fail to compile. Please verify this test actually compiles and passes.

e2e/src/test/scala/LazyStringFieldsSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+  "Lazy dictionary fields" should "work correctly through serialization" in {
+    import scalapb.LazyEncoder
+    val originalStringToInt = Map("hello" -> 1, "world" -> 2)
+    val originalIntToString = Map(1 -> "hello", 2 -> "world")
+    val original = LazyDictionary(stringToInt = originalStringToInt, intToString = originalIntToString)
+    
+    val bytes = original.toByteArray
+    val parsed = LazyDictionary.parseFrom(bytes)
+
+    parsed.stringToInt should contain theSameElementsAs originalStringToInt
```
thesamet, line 79: With file-level `lazy_string_fields: true`, the map entry messages for `LazyDictionary` also inherit the setting (via `message.getFile.scalaOptions.getLazyStringFields`). This means `stringToInt` would be typed as `Map[LazyField[String], Int]` instead of `Map[String, Int]`. Map lookups by `String` key would silently fail at runtime because `HashMap` calls `String.equals(LazyField[String])` on the stored keys, which returns `false`.

Consider whether `lazyStringFields` should be excluded from map entry key fields, and add a test that does `.get("hello")` on such a map to catch this.

benchmarks/results/LargeStringMessage__java.json
```diff
@@ -0,0 +1,316 @@
+[
```
thesamet: Please don't commit benchmark result JSON files to the main branch. They bloat the repository (4 files × ~316 lines each) and will create churn every time benchmarks are re-run. Consider a dedicated `benchmark-results` branch or linking to them from the PR description instead.

thesamet (commented):
Thanks for this PR — the performance gains are real and the approach of routing lazy string fields through `customSingleScalaTypeName` / TypeMapper is the right foundation. A few issues to address before this is ready to merge.

e2e/src/test/scala/LazyStringFieldsSpec.scala
poslegm: Yes, it compiles and passes with Scala 2 and Scala 3. I prove it with `sbt e2eJVM2_13/testOnly LazyStringFieldsSpec` and `sbt e2eJVM3/testOnly LazyStringFieldsSpec`. Conversions is defined [here](https://github.com/scalapb/ScalaPB/pull/2072/changes#diff-132e42af9bb3d33f4f09dde4aa117b93aa19b1540e3807724abe2784aa18db6eR14-R17) and [here](https://github.com/scalapb/ScalaPB/pull/2072/changes#diff-7384d44c34fedd2c9c784cb5e694a307a7a18606eb3ea13acf67bd762a53dde1R15-R21)

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
poslegm: `LazyEncoder` already used for implicit conversions: [example](https://github.com/scalapb/ScalaPB/pull/2072/changes#diff-7384d44c34fedd2c9c784cb5e694a307a7a18606eb3ea13acf67bd762a53dde1R11)

e2e/src/test/scala/LazyStringFieldsSpec.scala
poslegm: Lookups already works because of implicit conversions on `.get("hello")` calls. But you are right and this behaviour is error-prone because of untyped collection problem.

I excluded map keys from lazy parsing. Now parsing with `lazy_string_fields: true` results in `Map[String, T]` for `map<string, t>` proto dictionaries.

compiler-plugin/src/main/scala/scalapb/compiler/ProtobufGenerator.scala
poslegm: Wow, it was a serious bug that is difficult to detect, thank you for hint. Fixed it in commit https://github.com/scalapb/ScalaPB/pull/2072/changes/f4d5a12bc7e2c43632996306e70162abf5f79f92. Also I added e2eJson spec with proto←→json roundtrip using scalapb-json4s. It looks helpful for catching this sort of bugs. scalapb-json4s is connected as a dependency only to the `e2eJson` module, so it will not be a transitive dependency for scalapb.

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
poslegm: done, documentation clarified

scalapb-runtime/src/main/scala/scalapb/LazyField.scala
poslegm: I'm afraid that this behavior will be error-prone. If the user has decided to log the value of the field, it means that they are using this field, which means that the parsing is quite valid. Otherwise, the behavior will not be consistent:
```scala
def logName(field: String): Unit = logger.debug(s"Name is ${field}")

logName(proto.name) // Name is Joe
logger.debug(s"Name is ${proto.name}") // Name is LazyField(...)
```

It looks like pain for refactoring. On other hand, there is nothing wrong if parsing occurs in the case of accessing the field (or even logging the entire message at once). If the user wants to avoid parsing, they can simply skip logging the extra fields.

poslegm:
Hello @thesamet! Kindly reminder about this PR. Your feedback has been processed

PR #2112, opened 2026-05-20
Fix compilerplugin artifact for sbt 2

xuwei-k:


ybasket:
@thesamet Would you mind cutting a new alpha release that includes this fix? 🙏 It would really help as without the fix, sbt 2.0 + sbt-protoc is unusable.

thesamet:
v1.0.0-alpha.5 is being built and should be ready soon.

